PR #10, opened 2015-07-27
Fix README

amccloud:
No default is exported so you must destructure


mickhansen:
Right you are! Thank you for that :)
The documentation is very lacking, been busy just hacking some helpfull features together to make it possible to actually serve a decent amount of graphql queries.

PR #15, opened 2015-09-14
Add support for ARRAY and ENUM in TypeMapper

Aweary:
This should successfully translate `ARRAY` and `ENUM` sequelize types into the corresponding `GraphQLList` and `GraphQLEnumType`, respectively.

I also cleaned it up a little by using deconstructive assignment for `sequelizeTypes` and removing those `else` statements, since all `if` branches return the function when hit.

### DIFF
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -36,24 +36,29 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
   if (sequelizeType instanceof STRING ||
       sequelizeType instanceof TEXT ||
       sequelizeType instanceof UUID ||
-      sequelizeType instanceof DATE
-  ) return GraphQLString;
+      sequelizeType instanceof DATE) {
+        return GraphQLString
+      }
 
   if (sequelizeType instanceof ARRAY) {
     let elementType = toGraphQL(sequelizeType.type, sequelizeTypes);
-    return GraphQLList(elementType);
+    return new GraphQLList(elementType);
   }
 
   if (sequelizeType instanceof ENUM) {
     return new GraphQLEnumType({
-    values: sequelizeType.values.reduce((obj, value) => {
-      obj[value] = {value};
-      return obj;
-    }, {})});
+      values: sequelizeType.values.reduce((obj, value) => {
+        obj[value] = {value};
+        return obj;
+      }, {})
+    });
   }
 
   if (sequelizeType instanceof VIRTUAL) {
-    return toGraphQL(sequelizeType.returnType, sequelizeTypes);
+    let returnType = sequelizeType.returnType
+        ? toGraphQL(sequelizeType.returnType, sequelizeTypes)
+        : GraphQLString;
+    return returnType;
   }
 
   throw new Error(`Unable to convert ${sequelizeType.key || sequelizeType.toSql()} to a GraphQL type`);
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -1,26 +1,61 @@
-import { GraphQLInt, GraphQLString, GraphQLBoolean, GraphQLFloat } from 'graphql';
+import {
+  GraphQLInt,
+  GraphQLString,
+   GraphQLBoolean,
+   GraphQLFloat,
+   GraphQLEnumType,
+   GraphQLList
+ } from 'graphql';
 
+/**
+ * Checks the type of the sequelize data type and
+ * returns the corresponding type in GraphQL
+ * @param  {Object} sequelizeType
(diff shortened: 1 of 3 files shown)


src/typeMapper.js
```diff
@@ -1,26 +1,61 @@
...
+
+  if (sequelizeType instanceof BOOLEAN) return GraphQLBoolean;
+  if (sequelizeType instanceof FLOAT) return GraphQLFloat;
+  if (sequelizeType instanceof INTEGER) return GraphQLInt;
+
+  if (sequelizeType instanceof STRING ||
+      sequelizeType instanceof TEXT ||
+      sequelizeType instanceof UUID ||
+      sequelizeType instanceof DATE
+  ) return GraphQLString;
```
mickhansen: Lets do a `{}` here just because it's multline, otherwise i actually prefer one line short circuit if's (not sure why i went with else in my original code).

src/typeMapper.js
```diff
@@ -1,26 +1,61 @@
...
+
+  if (sequelizeType instanceof ARRAY) {
+    let elementType = toGraphQL(sequelizeType.type, sequelizeTypes);
```
mickhansen: could perhaps skip the variable and just pass `toGraphQL` directly to GraphQLList?

src/typeMapper.js
```diff
@@ -1,26 +1,61 @@
...
+    return GraphQLList(elementType);
```
mickhansen: Needs a `new` does it not? https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsSchema.js#L120

src/typeMapper.js
```diff
@@ -1,26 +1,61 @@
...
+    return new GraphQLEnumType({
+    values: sequelizeType.values.reduce((obj, value) => {
+      obj[value] = {value};
+      return obj;
+    }, {})});
+  }
+
+  if (sequelizeType instanceof VIRTUAL) {
+    return toGraphQL(sequelizeType.returnType, sequelizeTypes);
   }
```
mickhansen: A `returnType` might not be defined, should still fallback to string imo.

src/typeMapper.js
```diff
@@ -1,26 +1,61 @@
...
+    let elementType = toGraphQL(sequelizeType.type, sequelizeTypes);
+    return GraphQLList(elementType);
+  }
+
+  if (sequelizeType instanceof ENUM) {
+    return new GraphQLEnumType({
+    values: sequelizeType.values.reduce((obj, value) => {
+      obj[value] = {value};
+      return obj;
+    }, {})});
```
mickhansen: Some wonky identation here.

mickhansen:
I know we don't have specific tests for the type mapper (although we probably should at this point), so could you ammend https://github.com/mickhansen/graphql-sequelize/blob/master/test/attributeFields.test.js to cover the added `ARRAY`/`ENUM` support?

src/typeMapper.js
Aweary: Well that would leave us with a pretty long line, and I think it's better for readability. Using `elementType` its easy to see that its checking the type of the `ARRAY` element(s).

But it's ultimately up to you.

Aweary:
Should we maybe just add coverage for `typeMapper.js` then as well? Might as well address it now instead of putting it off for later.

src/typeMapper.js
mickhansen: Fine by me :)

mickhansen:
If you're up for doing it i think it's a great idea :)

src/typeMapper.js
Aweary: Would you be cool using a ternary operator for this? It's usually my preference when the if statement requires brackets (to avoid nested brackets) but I want to run it by you before committing.


This
```js
  if (sequelizeType instanceof VIRTUAL) {
    let returnType = sequelizeType.returnType
        ? toGraphQL(sequelizeType.returnType, sequelizeTypes)
        : GraphQLString;
    return returnType;
  }
```
Instead of
```js
  if (sequelizeType instanceof VIRTUAL) {
   if (sequelizeType.returnType) {
     return toGraphQL(sequelizeType.returnType, sequelizeTypes)
    }
  return GraphQLString
``

Aweary:
I added the `typeMapper.test.js` file for now, just to set it up for later. Once I get a bit more free time I can go back and add all the tests.

mickhansen:
Think the tests might be leaking into eachother.

mickhansen:
published as `v0.8.2`

PR #17, opened 2015-09-17
Added BIGINT type to mapper

jarno-steeman:


jarno-steeman:
In favor of this (too hasty made) PR https://github.com/mickhansen/graphql-sequelize/pull/16

src/typeMapper.js
```diff
@@ -31,7 +32,11 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
 
   if (sequelizeType instanceof BOOLEAN) return GraphQLBoolean;
   if (sequelizeType instanceof FLOAT) return GraphQLFloat;
-  if (sequelizeType instanceof INTEGER) return GraphQLInt;
+
+  if (sequelizeType instanceof INTEGER ||
+      sequelizeType instanceof BIGINT) {
```
mickhansen: what about the issue with `BIGINT` sometimes being too large to express as a javascript Number?

jarno-steeman:
It will not break since GraphQLInt has a check for this, but will return null if it's too large for JS to handle (safely). I'd say the only sollution would be to make a custom GraphQL type for this, but it would have to work with strings.

mickhansen:
Released as `v0.8.3`

PR #18, opened 2015-09-18
Add support for GraphQL-Relay Connections

### LINKED ISSUE #14: Support for connections
## I have the following connection setup on the user object, 

```
miners: {
      type: minerConnection,
      description: 'A users miners',
      args: connectionArgs,
      resolve:  (_, args) => resolver(models.Miner, args),
    },
```

this gives the the following error

```
Server request for query `AppHomeRoute` failed for the following reasons:

1. Cannot return null for non-nullable field MinerConnection.pageInfo.
   de{id,name},cursor},pageInfo{hasNextPage,hasPreviousPage}},i
                       
```

Is this designed to work with the :first and :find and :limit args that are needed for the graphql edge connection spec?

brad-decker:
### Included

* nodeInterface generator helper, which builds node definitions using a custom idFetcher and typeResolver. Requires the instance of sequelize and an object mapping models to GraphQL types. Example in test/relay.test.js 
* boolean check tool to determine if a given type is a connection by checking if it ends with the string 'Connection' which is what the connectionDefinitions function of graphql-relay-js names connection types (type + 'Connection')
* Adds two isConnection checks to core files, the first is in resolver. If a connection is detected the resolver will return the result of connectionsFromArray() passing in the field results in question and the args for the query. The second is in generateIncludes, which adds the fields passed in the edges clause of a query into the includes for the association.
* Includes a new test file relay.test.js

Implements and fixes #14


package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "babel-runtime": "^5.6.20",
+    "graphql-relay": "^0.3.2",
```
mickhansen: Should just be a devDependency right?

src/relay.js
```diff
@@ -0,0 +1,30 @@
...
+    if (models.any(model => model === type)) {
+      return sequelize.models[type].findById(id);
+    }
+    return null;
+  };
+}
+
+export function typeResolver(types) {
+  return obj => {
+    return types[obj.__options.name.singular];
```
mickhansen: `__options` is outdated, please use `obj.Model.options`.

test/relay.test.js
```diff
@@ -0,0 +1,236 @@
...
+    User.Tasks = User.hasMany(Task, {as: 'tasks'});
+
+    var node = sequelizeNodeInterface(sequelize, {User: userType, Task: taskType});
+    nodeInterface = node.nodeInterface;
+    nodeField = node.nodeField;
+
+    taskType = new GraphQLObjectType({
+      name: 'Task',
+      fields: {
+        id: globalIdField('Task'),
```
mickhansen: I was considering whether or not this should be automated for `graphql-sequelize` attributeFields when a user activates a `relay` option or similar.

test/relay.test.js
```diff
@@ -0,0 +1,236 @@
...
+      name: 'User',
+      fields: {
+        id: globalIdField('User'),
+        name: {
+          type: GraphQLString
+        },
+        tasks: {
+          type: taskConnection,
+          args: connectionArgs,
+          resolve: resolver(User.Tasks)
```
mickhansen: Does this also support stuff like `totalCount` and similar?
Would the resolver actually need to be further down on a `edges` field to support that?

mickhansen:
Looks good initially. Just a few exploratory comments since i'm not too familiar with the relay-way yet.
I'd love to clarify how to handle extra meta fields for a field, i.e. totalCount, and where would args go, on `tasks` or on `edges`?

package.json
brad-decker: Absolutely. will fix.

test/relay.test.js
brad-decker: This has to be done globally once all types are defined as it sets up a typeResolver idFetcher function.

test/relay.test.js
brad-decker: It supports the default relay connection args

```js
export var connectionArgs: GraphQLFieldConfigArgumentMap = {
  before: {
    type: GraphQLString
  },
  after: {
    type: GraphQLString
  },
  first: {
    type: GraphQLInt
  },
  last: {
    type: GraphQLInt
  },
};
```

connections implement connectionType as follows:

```js

/**
 * The common page info type used by all connections.
 */
var pageInfoType = new GraphQLObjectType({
  name: 'PageInfo',
  description: 'Information about pagination in a connection.',
  fields: () => ({
    hasNextPage: {
      type: new GraphQLNonNull(GraphQLBoolean),
      description: 'When paginating forwards, are there more items?'
    },
    hasPreviousPage: {
      type: new GraphQLNonNull(GraphQLBoolean),
      description: 'When paginating backwards, are there more items?'
    },
    startCursor: {
      type: GraphQLString,
      description: 'When paginating backwards, the cursor to continue.'
    },
    endCursor: {
      type: GraphQLString,
      description: 'When paginating forwards, the cursor to continue.'
    },
  })
});

var edgeType = new GraphQLObjectType({
    name: name + 'Edge',
    description: 'An edge in a connection.',
    fields: () => ({
      node: {
        type: nodeType,
        description: 'The item at the end of the edge',
      },
      cursor: {
        type: new GraphQLNonNull(GraphQLString),
        description: 'A cursor for use in pagination'
      },
      ...resolveMaybeThunk(edgeFields)
    }),
  });

  var connectionType = new GraphQLObjectType({
    name: name + 'Connection',
    description: 'A connection to a list of items.',
    fields: () => ({
      pageInfo: {
        type: new GraphQLNonNull(pageInfoType),
        description: 'Information to aid in pagination.'
      },
      edges: {
        type: new GraphQLList(edgeType),
        description: 'Information to aid in pagination.'
      },
      ...resolveMaybeThunk(connectionFields)
    }),
  });
```

The resolver for these custom fields are handled by the graphql-relay packages but I should add some tests for these to make sure all this functionality works.

brad-decker: Looks like GraphQL-relay leaves it up to the user to supply totalCount field. From the tests:

```js
var {connectionType: friendConnection} = connectionDefinitions({
  name: 'Friend',
  nodeType: userType,
  edgeFields: () => ({
    friendshipTime: {
      type: GraphQLString,
      resolve: () => 'Yesterday'
    }
  }),
  connectionFields: () => ({
    totalCount: {
      type: GraphQLInt,
      resolve: () => allUsers.length
    }
  }),
});
```

I can easily implement this with a connectionDefinitions helper for sequelize that adds totalCount by default?

test/relay.test.js
mickhansen: Alright!

test/relay.test.js
mickhansen: Should we bake in support for the default connection args then? before/after should be trivial i suppose, first/last is more tricky since it depends on what field should be ordered on (could be createdAt by default but then needs a syntax for changing that).

That graphql-relay test does not look to match what we have in terms of having the resolver on the top level `tasks`? Would one need to have the resolver on `edges` generally to supply extra fields?

test/relay.test.js
```diff
@@ -0,0 +1,236 @@
...
+      });
+    });
+  });
+
+  it('should resolve a plain result with a single model', function () {
+    var users = this.users;
+
+    return graphql(schema, `
+      {
+        users {
```
mickhansen: would it be prudent to make users a relay connection type aswell? Just to test that nested connections work as intended, possibly with a single query using joins?

package.json
mickhansen: Yeah i saw that, ideally i'd like to avoid that, if the user is using relay they'll add it and we should still be able to require it, we just need some way of detecting when to include relay stuff.

test/relay.test.js
brad-decker: Yeah thats a good idea. I'll get started on this.

test/relay.test.js
brad-decker: We could add support to the sequelize queries to enhance improvement of queries, yes. Resolvers  don't go onto the edges -> Graphql-relay handles all of this magic for us. The added lines in the generateIncludes handles mapping fields from the nodes to the include calls to grab data.

package.json
brad-decker: So are you suggesting we move it to dev dependencies and if the user tries to require from the relay file they will get an error? I don't think this is the best case and graphql-relay is a requirement for the work we've done thus far.

test/relay.test.js
brad-decker: Had an issue where nested connections were being added to attributes inside of include generator. Fixed this by checking for the existence of an edges field on each field passed in the simpleAST.

package.json
mickhansen: Something like that possible yeah, i've just been happy with being able to not depend on `relay` directly and let the user define the version he needs to a certain degree.

I'd love this lib to be a helper when needed than something you have to use and customize for all fields (not sure if that would be an issue with the graphql-relay package actually).

A reason to not use own graphql package is also that stuff like `instanceof GraphQLList` would fail since the constructor would differ between two packages - Might not be an issue with relay.

Lets give having it as a straight dependency a go and possible test if it works when the user has his own version installed aswell.

test/relay.test.js
```diff
@@ -0,0 +1,323 @@
...
+        type: Sequelize.STRING
+      }
+    }, {
+      timestamps: false
+    });
+
+    User.Tasks = User.hasMany(Task, {as: 'tasks'});
+    Project.Users = Project.hasMany(User, {as: 'users'});
+
+    var node = sequelizeNodeInterface(sequelize, {Project: projectType, User: userType, Task: taskType});
```
mickhansen: Wouldn't the types be undefined at this point?

test/relay.test.js
mickhansen: Sorry for being a bit slow, i'm getting it now :) People will manually generate connectionDefinitions with types so can easily add extra fields themselves.

mickhansen:
Just a single comment with undefined variables - I'll do a preliminary merge after that, this feature obviously still needs a lot more work but lets start playing around with it in real code :)

test/relay.test.js
Aweary: `nodeInterface`s would be using those types by reference to return them when querying for an item by its `id`. Since it doesn't actually call them until at all (only returns the reference) it does not matter. We've verified these tests all pass as well.

brad-decker: Agreed with @Aweary and I don't see another alternative to this approach other than forcing types to be named a certain way. So if the Sequelize model is named User, then the type has to be named userType.

I think its a more robust approach to explicitly declare the relationship as we are doing here.

brad-decker:
We'll be using this setup in our real code base and testing intensely, any issues we uncover along the way we will submit fixes for.

test/relay.test.js
mickhansen: `projectType`, `userType` and `taskType` are all defined after the `sequelizeNodeInterface` call, so `typeResolver` should be getting an object with keys pointing at `undefined` values?

There must be something i'm missing, if the values are not required, the test (which will serve as documentation) would be better off with explicitely undefined values.

mickhansen:
@brad-decker sounds good, hoping to integrate it into my project soon aswell - as long as graphql/relay is in flux, so will this project, will have to live with that :)

test/relay.test.js
brad-decker: the declarations of the types are hoisted to the top of the file and at that time they are set in memory as undefined. The references to the memory allocations are passed along in the object passed to the sequelizeNodeInterface function. However by the time the type resolver is ran, which is ran when a query comes into the graphql server, the types are defined in memory and thus there is no issue with undefined variables.

brad-decker: The object is necessary, unfortunately. sequelizeNodeInterface tells graphql how to determine which graphql type to association with an object, and the only way to do that is by mapping Model names to the types that are defined for them.

mickhansen: @brad-decker Javascript does not work quite like that, only thing in javascript passed by reference is objects and arrays, so for this case to work we'd have to save the map in an array and then set the types via that.

The code is only used as an interface in the test, so the typeResolver code is actually never called. If you pass an empty object to `sequelizeNodeInterface` it works just aswell.

So something is not quite obvious here.

brad-decker: The typeResolver is ran when relay runs a node query, will expand upon the test to query a node to prove its effectiveness.

mickhansen: We're not using the returned `nodeField`, which would probably fail were it actually being run.

mickhansen: @brad-decker Please do, i'm fairly certain the code will fail in that case - So help in clearing up my confusion would be great :)

mickhansen: `types` needs to be an object that is later manipulated by the user.

I propose something like:

```
let {
   nodeInterface,
   nodeField,
   nodeTypes
} = sequelizeNodeInterface(sequelize);

nodeTypes.Task = taskType;
```

mickhansen: Or we simply remove `sequelizeNodeInterface` and simply provide the idFetcher.

src/relay.js
```diff
@@ -0,0 +1,30 @@
+import {fromGlobalId, connectionFromArray, nodeDefinitions} from 'graphql-relay';
+
+export function idFetcher(sequelize) {
+  return globalId => {
+    const {type, id} = fromGlobalId(globalId);
+    const models = sequelize.models.map(model => model.name);
```
mickhansen: `sequelize.models` is an object/map.

brad-decker: Working through some of these bugs while i'm testing the node queries. will submit some changes shortly .

mickhansen: Great, thanks for your work & effort @brad-decker

test/relay.test.js
brad-decker: You're absolutely right about the undefined types. so this is how the operation works when you make a node query:

idFetcher is called with the global id. The id fetcher breaks it a part into its constituents and then does a query to get the object. The object is passed into type Resolver, which then needs to get t he GraphQL type to mold the data into. 

I committed and pushed a solution which is to do away with the automatic type resolver, but still use the idFetcher magic. But this solution seems ugly to me. @Aweary is looking into it as well so any input you provide is valued.

src/relay.js
```diff
@@ -0,0 +1,32 @@
+import {fromGlobalId, connectionFromArray, nodeDefinitions} from 'graphql-relay';
+
+export function idFetcher(sequelize) {
+  return globalId => {
+    let {type, id} = fromGlobalId(globalId);
+    const models = Object.keys(sequelize.models);
+    type = type.toLowerCase();
```
mickhansen: why the `toLowerCase`?

test/relay.test.js
```diff
@@ -0,0 +1,350 @@
...
+        case 'project':
+          return projectType;
+          break;
+        case 'task':
+          return taskType;
+          break;
+        default:
+          return null;
+      }
+    });
```
mickhansen: `sequelizeNodeInterface` could also return a `typeMap` property which the user needs to fill out with `modelName => type` mappings.

```js
let {
  nodeInterface,
  nodeField,
  nodeTyperMapper
} = sequelizeNodeInterface(sequelize);

nodeTypeMapper[User.name] = userType;
```

With the user providing a resolve function themselves they could just as well do:

```
nodeDefinitions(idFetcher, (obj) => { ... });
```

Providing a helper in that case seems a little pointless.

brad-decker: that seemed to work perfectly. Going to submit the changes shortly.

src/relay.js
brad-decker: The reason we need it is because of the possibility of defining types like what we did:

```
id: globalIdField('Task'),
```

Just to make sure that in the event someone types the models with upper case name its forced to lower case to test for equality in the models object

mickhansen: In that case the model names should be cast to lower case aswell, otherwise you're matching possible uppercase model names against a forced lowercase type.

mickhansen:
Merging this for now, let's play with it in real world settings and see if it breaks.
@brad-decker the nodeField/nodeTypeMapper stuff is really only for queries like:

```
{
  node(id: 1) {
    id
    
    ... on User {
   
    }
  }
}
```

Which i don't personally believe i'll need in my frontend code for the next good while.

mickhansen:
Published in `v0.9.0`

brad-decker:
@mickhansen Yeah the node queries are for relay to re-query objects for updates without knowing the objects type.

src/generateIncludes.js
```diff
@@ -96,5 +97,15 @@ export default function generateIncludes(simpleAST, type, root, options) {
...
+  if (isConnection(type)) {
+    let node = simpleAST.fields.edges.fields.node;
+    let fields = [];
+    _.forIn(node.fields, (field, key) => {
+      if (!field.fields.hasOwnProperty('edges')) {
+        fields.push(key);
+      }
+    });
+    result.attributes = result.attributes.concat(fields);
+  }
```
mickhansen: Just FYI this code was removed with https://github.com/mickhansen/graphql-sequelize/commit/26c1955b8a30f78cb7cab827735b1420fdc71899

We now support proper nested includes with nested connections.

PR #23, opened 2015-09-22
Remove Special characters from values object keys

### LINKED ISSUE #22: Enum Value generator names are not accepted by GraphQL
Enum Value object names can't contain spaces, special characters etc. Already have a fix ready to push but logging the error for version history etc.

brad-decker:
Fixes #22 and provides better overall support for Enum Value object

### DIFF
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -29,6 +29,9 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
     VIRTUAL
   } = sequelizeTypes;
 
+  // Regex for finding special characters
+  const specialChars = /[^a-z\d]/i;
+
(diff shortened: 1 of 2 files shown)


mickhansen:
How does graphql-relay implement first/last and before/after?
If it's not express in SQL is it simply filtering after receiving a list (because that would be suboptimal)

brad-decker:
Yeah, deeper native integration would be great. I'll put it on my todo list but for testing purposes we can play around with it for now. I'd like to see tests also with large data sets using a real data store and not sqlite.

mickhansen:
Yeah i have on my todo to convert the tests to use docker-compose with postgres rather than sqlite (since sqlite doesn't have all the features that Sequelize does).

PR #25, opened 2015-09-22
Add support for GraphQL Fragments

### LINKED ISSUE #24: Support GraphQL fragments
Currently `graphql-sequelize` will fail if a query contains a fragment. For example: 

```js
query RootQueryType { post(id: 1) { title, ...contentFrag }} fragment contentFrag on Post { content }
```

Will initially fail with an error:

```js

TypeError: Cannot read property 'reduce' of undefined
    at selections.reduce.fields (/Users/[redacted]/API/node_modules/graphql-sequelize/lib/simplifyAST.js:44:53)
    at Array.reduce (native)
    at simplyAST (/Users/[redacted]/node_modules/graphql-sequelize/lib/simplifyAST.js:30:21)
    at resolver (/Users/[redacted]/node_modules/graphql-sequelize/lib/resolver.js:63:50)
    at resolveField (/Users/[redacted]/node_modules/graphql/execution/execute.js:399:18)
```

Because:

```js
  simpleAST.fields[key].args = selection.arguments.reduce(function (args, arg) {
      args[arg.name.value] = arg.value.value;
      return args;
    }, {});
```

Assumes that a field will have an arguments property, while `fragmentSpread` will not.

brad-decker:
As stated in #24 the simplifyAST function was failing on fragment queries. Adding support was easy as long as we have access to the info object which contains a fragments object containing all fragments on the query. Fixes #24

### DIFF
--- a/src/attributeFields.js
+++ b/src/attributeFields.js
@@ -1,5 +1,5 @@
 import * as typeMapper from './typeMapper';
-import { GraphQLNonNull } from 'graphql';
+import { GraphQLNonNull, GraphQLEnumType } from 'graphql';
 
 module.exports = function (Model, options) {
   options = options || {};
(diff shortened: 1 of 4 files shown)


mickhansen:
Going to merge this for now. I'll probably look into an AST only solution later.

mickhansen:
`v0.10.0`

PR #28, opened 2015-09-29
expose Relay API

Aweary:
@mickhansen just a super quick change so these Relay functions can be used.


mickhansen:
That's probably a good idea.

mickhansen:
`v0.10.3

Aweary:
haha, thank you! :smile:

Issue #30, opened 2015-09-29
idFetcher only supports sequelize models


Aweary:
So looking at `idFetcher`

```js
export function idFetcher(sequelize) {
  return globalId => {
    let {type, id} = fromGlobalId(globalId);
    const models = Object.keys(sequelize.models);
    type = type.toLowerCase();
    if (models.some(model => model === type)) {
      return sequelize.models[type].findById(id);
    }
    return null;
  };
}
```

I've found that there's no way to fetch a GraphQL type that doesn't have an associated sequelize model. This is bad since Relay apps often use a pattern where a root `viewer` field points to all the other actual fields, making it a sort of virtual field that just acts as an access point.

mickhansen:
We could provide a `ifNotModelCallback` or similar.
Wouldnt a root `viewer` be without an id? I use a root viewer in my current graphql setup and just return the current authenticated user.

Aweary:
You just need to return an object with an `id` in the `viewer` resolve function. Currently this setup doesn't have any notion of users, so it just needs a  `viewer` field as an access point, since you can't have arrays as root nodes currently.

It totally works right now, but the issue is that Relay issues a `node` query with the ID when using `setVariables` instead of just calling `viewer` and the `idFetcher` is returning null.

I think maybe we can adjust `typeMapper` to allow for users to declare a type name and type without an associated model.

mickhansen:
Ah hmm, can you link me to some relevant relay documentation on how they handler `viewer`?
As far as i can tell the user couldn't you just supply a full resolver to viewer instead of a partial one?
Or is it a partial resolve so it does not need to load the viewer unless specifically asked?

Aweary:
`viewer` is non-standard and not part of the Relay spec, it's just a regular field. It's a workaround while root nodes are being worked on https://github.com/facebook/relay/issues/112

`viewer` just resolves to an object of type `viewerType` which has all main types as nullable fields.

Aweary:
This all works well and fine, loading data is perfect. But Relay uses the `nodeInterface` when doing secondary queries (like updating a variable with `this.props.relay.setVariables`) so it passes the ID of `viewer` (which is the base64 encoding of `'viewer:1'`. `idFetcher` is invoked and it returns null because theres no `viewer` in the `models` list.

mickhansen:
Allowing extensions to the type mapper would probably be a good idea in any case.

mickhansen:
[...]

PR #31, opened 2015-09-29
add support for unassociated GraphQL types

Aweary:
This should do the trick. All you need to do is use `typeMapper` to associate the name and GraphQL type and it will correctly resolve it when using the `nodeInterface` and `nodeField` methods

Resolves https://github.com/mickhansen/graphql-sequelize/issues/30

### DIFF
--- a/test/relay.test.js
+++ b/test/relay.test.js
@@ -50,6 +50,7 @@ describe('relay', function () {
     , nodeInterface
     , Project
     , projectType
+    , viewerType
     , userConnection
     , nodeField
     , schema;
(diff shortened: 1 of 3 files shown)


test/relay.test.js
```diff
@@ -136,10 +137,24 @@ describe('relay', function () {
...
+      }),
+      interfaces: [nodeInterface]
+});
+
     nodeTypeMapper.mapTypes({
       [User.name]: userType,
       [Project.name]: projectType,
-      [Task.name]: taskType
+      [Task.name]: taskType,
+      viewer: viewerType
```
mickhansen: I'm not totally happy with the uncased matching, but i doubt people willa ctually have types with different cases of the same name - But if it does happen we'll break it later.

Aweary: Yeah, it's already normalizing the types here https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L25 but I think its safe for right now. Not a huge issue changing that.

Aweary:
@brad-decker is currently out of the office for the week, so he may be slow to respond on this. Just a heads up.

mickhansen:
Alright, would like his feedback since he implemented it originally - How urgently do you need it merged?

Aweary:
Not a big rush, I'm just pulling `graphql-sequelize` from our fork instead of `npm`, so we can wait it out. :+1:

mickhansen:
@brad-decker Back yet? :)

Aweary:
@brad-decker is back, and we discussed the casing coercion and decided that we should just remove it. So model/type names are case sensitive.

brad-decker:
Sorry for the delay @mickhansen

mickhansen:
@brad-decker Not a problem, just wanted to ping :)

Aweary:
:tada: :tada: :tada:

mickhansen:
`v0.11.0`

PR #35, opened 2015-10-08
Better support for enums with dashes

idris:
Before, GraphQL would throw an error.
Now it works closer to what’s expected.
The value in GraphQL is still camelCased, but now it resolves properly.

### DIFF
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -58,8 +58,9 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
   if (sequelizeType instanceof ENUM) {
     return new GraphQLEnumType({
       values: sequelizeType.values.reduce((obj, value) => {
+        let sanitizedValue = value;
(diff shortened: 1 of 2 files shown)


mickhansen:
Would love to have a test for it so we don't have a regression.

idris:
Alright, test is up. What's the best place to document the camelCasing behavior with enums?

mickhansen:
In the field helpers section i think.

idris:
@mickhansen okay, I've got a commit ready with documentation, but had a quick question/suggestion:

GraphQL [supports underscores, and recommends all-caps + underscores](https://facebook.github.io/graphql/#sec-Enum-Value), so what do you think about converting special characters to underscores instead of camelCase.  It should also simplify the code a little bit.

mickhansen:
Not sure actually, mapping between GraphQL conventions and Postgres contentions might be tricky.
I don't use enums so don't really have a strong opinion.

PR #50, opened 2015-10-28
Use primaryKeyAttribute to resolve globalIdField

idris:
The default will use `instance.id`, which is no good when we have a different `id`

### DIFF
--- a/src/attributeFields.js
+++ b/src/attributeFields.js
@@ -28,7 +28,7 @@ module.exports = function (Model, options) {
   }, {});
 
   if (options.globalId) {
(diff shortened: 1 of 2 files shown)


idris:
Not sure why those tests are failing..  they pass on my local

mickhansen:
Unsure why they are failing either, i'll look into it.

mickhansen:
Passing now, can you add a unit test for non-id primary key models though?

mickhansen:
ping @idris

idris:
Done!

PR #104, opened 2015-12-20
Connections for no association.

ivnivnch:
Users connection is not associated with the viewer.
It can be used to query all users:
```
const {User} = sequelize.models;

const GraphQLUser = new GraphQLObjectType({
  name: 'User',
  fields: () => attributeFields(User, {globalId: true}),
  interfaces: () => [nodeInterface]
});

const GraphQLViewer = new GraphQLObjectType({
  name: 'Viewer',
  fields: () => {
    const usersConnection = sequelizeConnection({
      name: 'Users',
      nodeType: GraphQLUser,
      target: User // !!! no association
      /* orderBy, where, connectionFields, ... */
    })
    
    return {
      id: globalIdField('Viewer'),
      users: {
        type: usersConnection.connectionType,
        args: usersConnection.connectionArgs,
        resolve: usersConnection.resolve
      }
    }
  }
});
```

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -100,7 +100,10 @@ module.exports = function (target, options) {
         });
       });
     }
-    return model[list ? 'findAll' : 'findOne'](findOptions).then(function (result) {
+    return model[list || isConnection(info.returnType) ? 'findAll' : 'findOne'](findOptions).then(function (result) {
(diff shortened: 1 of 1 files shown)


mickhansen:
Mind add a testing aswell?

PR #122, opened 2016-01-26
attributeFields - Map option

bringking:
This PR implements the map option for attributeFields with an object hash and function. Included are two unit tests that cover using a function or using an object hash

### DIFF
--- a/src/attributeFields.js
+++ b/src/attributeFields.js
@@ -12,26 +12,25 @@ module.exports = function (Model, options) {
     var attribute = Model.rawAttributes[key]
       , type = attribute.type;
 
-    // determine the key
-    var mappedKey = key;
+
     if (options.map) {
       if (typeof options.map === 'function') {
-        mappedKey = options.map(key);
+        key = options.map(key) || key;
       } else {
-        mappedKey = options.map[key] || key;
+        key = options.map[key] || key;
       }
     }
 
-    memo[mappedKey] = {
+    memo[key] = {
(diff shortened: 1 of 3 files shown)


src/attributeFields.js
```diff
@@ -12,16 +12,26 @@ module.exports = function (Model, options) {
     var attribute = Model.rawAttributes[key]
       , type = attribute.type;
 
-    memo[key] = {
+    // determine the key
+    var mappedKey = key;
+    if (options.map) {
+      if (typeof options.map === 'function') {
+        mappedKey = options.map(key);
```
mickhansen: Might want a `|| key` here too, let users ignore those they don't want to map

src/attributeFields.js
```diff
@@ -12,16 +12,26 @@ module.exports = function (Model, options) {
     var attribute = Model.rawAttributes[key]
       , type = attribute.type;
 
-    memo[key] = {
+    // determine the key
+    var mappedKey = key;
```
mickhansen: Couldn't we just overwrite key? I don't see it being used for anything else

bringking: Yeah I agree, not sure why I wrote it that way. I will fix it ASAP

bringking:
Updated based on your comments, let me know if it looks okay

mickhansen:
Looks good, can i ask you to add a documentation update aswell?

bringking:
Absolutely, give me a few mins

bringking:
FYI these docs are complete

PR #123, opened 2016-01-27
filterAttributes was defaulting to false all the time

bringking:
Set to a default arg of true, and fixed a relay test where it was expecting it to be false. Addresses #124

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -9,7 +9,7 @@ function inList(list, attribute) {
   return ~list.indexOf(attribute);
 }
 
-function resolverFactory(target, options) {
+function resolverFactory(target, options = {filterAttributes: true}) {
   var resolver
     , targetAttributes
     , isModel = !!target.getTableName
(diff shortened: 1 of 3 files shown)


bringking:
Hmm not sure why the travis build failed.. the integration tests were passing for me locally

src/resolver.js
```diff
@@ -24,7 +24,6 @@ function resolverFactory(target, options) {
   if (options.before === undefined) options.before = (options) => options;
   if (options.after === undefined) options.after = (result) => result;
   if (options.handleConnection === undefined) options.handleConnection = true;
-  if (options.filterAttributes) options.filterAttributes = resolverFactory.filterAttributes;
```
mickhansen: `resolverFactory.filterAttributes` should be changed to true, not sure how i made that mistake.
Setting a default options would not work for everyone who has options defined.

bringking: Okay I made this change

PR #125, opened 2016-01-28
GraphQLEnumType name

ivnivnch:


ivnivnch:
I use nodeTypeMapper.mapTypes for my Sequelize models:
```
sequelize.define('First', {
  type: Sequelize.ARRAY(Sequelize.ENUM('a', 'b', 'c'))
});

sequelize.define('Second', {
  type: Sequelize.ARRAY(Sequelize.ENUM('a', 'b', 'c'))
});
```
Error: Schema must contain unique named types but contains multiple types named "undefined".

After adding a name:
```
sequelize.define('First', {
  type: Sequelize.ARRAY(Sequelize.ENUM({name: 'first_type', values: ['a', 'b', 'c']}))
});

sequelize.define('Second', {
  type: Sequelize.ARRAY(Sequelize.ENUM({name: 'second_type', values: ['a', 'b', 'c']}))
});
```

mickhansen:
Hmm, in my opinion this should probably be moved to attributeFields, let it use the name of the attribute as the name of the ENUM

/cc @brad-decker

ivnivnch:
Ok. It will be better this way. Also Array of Enum currently is not supported.

Issue #126, opened 2016-01-28
Map known String attributes in typeMapper


bringking:
I want to get your thoughts on mapping some column types, that sequelize doesn't have a concrete type for. 

For example, on my project we are running into string column types in our models for things like -

'longblob'
'char(72)'

With things like 'longblob', we can exclude that value, but it would be nice to supply a custom typeMapper for things like 'char' that tells graphql-sequelize to make it a string type. Something like -

```js
let autoMapped = attributeFields(Models.SomeLegacyThingWithChar,{
  typeMap:{
    'char(72)': GraphQLString,
    'longblob': GraphQLString
  }
});
```
I can work on this, if you think it is a useful addition.

bringking:
hmm actually looking at the current sequelize source, longblog and char are supported concrete types. Might be useful to still provide a way to override type mapping though

mickhansen:
For it to make sense it would probably be best if you could inject types at a higher level.
In the example you provided the user might aswell exclude htose two types and define them seperately.

bringking:
Right, I think that makes sense

mickhansen:
So maybe something like:

```js
import {typeMapper} from 'graphql-sequelize';
typeMapper.map([
  ['longblob', GraphQLString],
  [DataTypes.JSONB, GraphQLObjectType]
])
```

PR #137, opened 2016-02-02
Added GraphQLJSON type for JSON & JSONB DataTypes

ivnivnch:


bringking:
Looks like the map function tests are failing -

  1) attributeFields should be possible to rename fields with a object map:
  2) attributeFields should be possible to rename fields with a function that maps keys:
 
Also, IMHO it should be considered if this library should be responsible for providing custom GraphQL types, or just allow users to provide their own mappings via some method like discussed in #126. Interested to hear @mickhansen's thoughts

mickhansen:
@bringking Yeah they're failing on master aswell, not sure why

mickhansen:
I'm not sure about this, a custom JSON scalar might make sense. But most of the time users should probably supply their own custom ObjectType for a JSON field so you still have some sort of schema.

ivnivnch:
This can be useful if the shape of JSON object is changed at runtime. It can be defined in the scheme, but I see it as a helper object is the same as the other helper functions.

mickhansen:
Definitely, i like this as a possible default with the option of letting users owerwrite it both via attributeFields and type mapping.

ivnivnch:
This is possible with exclude/only options for attributeField function.

mickhansen:
I'm aware :)

mickhansen:
Custom JSON scalar was added by @bringking: https://github.com/mickhansen/graphql-sequelize/blob/master/src/types/jsonType.js

PR #143, opened 2016-02-03
added the ability to supply a custom mapType function to the typeMapper

### LINKED ISSUE #126: Map known String attributes in typeMapper

bringking:
This resolves #126 and gives users the ability to override the mapping of types with a custom function. Added tests to typeMapper.test.js and updated readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,51 @@ userType = new GraphQLObjectType({
   })
 });
 ```
+### Providing custom types
+
+`attributeFields` uses the graphql-sequelize `typeMapper` to map Sequelize types to GraphQL types. You can supply your own
+mapping function to override this behavior using the `mapType` export.
+
+```js
+var Model = sequelize.define('User', {
+  email: {
+    type: Sequelize.STRING,
+    allowNull: false
+  },
+  isValid: {
+    type: Sequelize.BOOLEAN,
+    allowNull: false
+  }
+});
+
+import {attributeFields,typeMapper} from 'graphql-sequelize';
+typeMapper.mapType((type) => {s
+   //map bools as strings
+   if (type instanceof Sequelize.BOOLEAN) {
+     return GraphQLString
+   }
+   //use default for everything else
+   return false
+});
+
+//map fields
+attributeFields(Model);
+
+/*
+{
+  id: {
+    type: new GraphQLNonNull(GraphQLInt)
+  },
+  email: {
+    type: new GraphQLNonNull(GraphQLString)
+  },
+  isValid: {
+      type: new GraphQLNonNull(GraphQLString)
+  },
+}
+*/
+
+```
 
 ### Renaming generated fields
 
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -1,11 +1,21 @@
 import {
-  GraphQLInt,
-  GraphQLString,
+   GraphQLInt,
+   GraphQLString,
    GraphQLBoolean,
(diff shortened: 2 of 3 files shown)


src/typeMapper.js
```diff
@@ -1,12 +1,22 @@
...
  } from 'graphql';
 
+let _userMappingFunc;
+/**
+ * A function to set a custom mapping of types
+ * @param mapFunc
+ */
+export const mapType = (mapFunc = () => false) => {
+    _userMappingFunc = mapFunc;
+};
```
mickhansen: Don't really need a default for the argument do we, if the user calls it they should really have a method :p

mickhansen:
Be great if you could convert your changes to double-space indentation

README.md
```diff
@@ -187,6 +187,51 @@ userType = new GraphQLObjectType({
   })
 });
 ```
+### Providing custom types
+
+```attributeFields``` uses the graphql-sequelize ```typeMapper``` to map Sequelize types to GraphQL types. You can supply your own
```
mickhansen: for single words you can simply use `attributeFields`

src/typeMapper.js
```diff
@@ -1,12 +1,22 @@
...
    GraphQLList
  } from 'graphql';
 
+let _userMappingFunc;
+/**
+ * A function to set a custom mapping of types
+ * @param mapFunc
+ */
+export const mapType = (mapFunc = () => false) => {
+    _userMappingFunc = mapFunc;
```
mickhansen: Maybe just `customTypeMapper`

bringking: yeah I like that better

README.md
bringking: ah good to know. Will fix

src/typeMapper.js
bringking: My only reasoning for this was lower in the file, when we use their function, we don't run into errors when we do `_userMappingFunc(sequelizeType)`. The alternative would be set the _userMappingFunc to a function in the declaration. I guess that still doesn't solve someone passing an object or something.. I can take it out and type check it before I call it to be extra sure.

mickhansen: No reason to silently fail if the user calls it with something like `mapType(Object)` or `mapType(undefined)`

bringking: Yeah I guess that is true, I can just take it out.

bringking:
Okay sorry for the churn, the latest push should incorporate all your feedback.

mickhansen:
Okay just two nitpicks! And if you could clean up the commit message just slightly: https://github.com/RinconStrategies/graphql-sequelize/commit/889e1da000cf6921af3b46867ef671113786d674 :)

bringking:
Okay changes made! Thanks for being patient, I will be better about adhering to your style guidelines in the future.

mickhansen:
Thanks :)

Issue #145, opened 2016-02-04
defaultArgs additions


bringking:
It would be great to have defaultArgs have the ability have a general "where" clause included for more advanced filtering. Something like-

```js
{
  users(where: {
    user_id:1,
    lockedEndDate:
      {
        $gte:"1/1/2016"
      }
    })
}
```

Not sure if it can be a GraphQLObject, or if it will have to be a string..

bringking:
Fixed with #150. Closing

PR #150, opened 2016-02-11
initial where implementation

bringking:
This PR implements a "where" arg on the defaultArgs and defaultListArgs. This implementation is slightly naive but does a few things -

* Implements a custom Scalar type called "JSONType", inspired by - https://github.com/graphql/graphql-js/pull/242

* Maps the operators defined [here](http://docs.sequelizejs.com/en/latest/docs/querying/) to add back the "$" since that isn't an allowed character in GraphQL InputTypes

* Makes the primary_key arg on defaultArgs nullable, since you don't need it if you are providing a "where" query.

I say that it is naive because it would be nice to map the where operators for each model to concrete types, but I struggled with this because of the possibility of nested "$and" and "$or" values in the "where" queries defined by sequelize. I was close to getting this implemented, but I think this is a wholly acceptable solution for the time being. 

This PR resolves issue #145

### DIFF
--- a/test/unit/replaceWhereOperators.test.js
+++ b/test/unit/replaceWhereOperators.test.js
@@ -1,0 +1,61 @@
+import chai, {expect} from "chai";
+import {replaceWhereOperators} from "../../src/replaceWhereOperators";
+
+describe("replaceWhereOperators", ()=> {
+  it("should take an Object of grapqhl-friendly keys and replace with the correct sequelize operators", ()=> {
+    let before = {
+      and: 1,
+      or: "1",
+      gt: [{and: "1", or: "1"}, {between: "1", overlap: "1"}],
+      gte: 1,
+      lt: {
+        and: {
+          "test": [{or: "1"}]
+        }
+      },
+      lte: 1,
+      ne: 1,
+      between: 1,
+      notBetween: 1,
+      in: 1,
+      notIn: 1,
+      notLike: 1,
+      iLike: 1,
+      notILike: 1,
+      like: 1,
+      overlap: 1,
+      contains: 1,
+      contained: 1,
+      any: 1,
+      col: 1
+    };
+    let after = {
+      $and: 1,
+      $or: "1",
+      $gt: [{$and: "1", $or: "1"}, {$between: "1", $overlap: "1"}],
+      $gte: 1,
+      $lt: {
+        $and: {
+          "test": [{$or: "1"}]
+        }
+      },
+      $lte: 1,
+      $ne: 1,
+      $between: 1,
+      $notBetween: 1,
+      $in: 1,
+      $notIn: 1,
+      $notLike: 1,
+      $iLike: 1,
+      $notILike: 1,
+      $like: 1,
+      $overlap: 1,
+      $contains: 1,
+      $contained: 1,
+      $any: 1,
+      $col: 1
+    };
+
+    expect(replaceWhereOperators(before)).to.deep.equal(after);
+  })
+});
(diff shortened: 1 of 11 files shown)


mickhansen:
Possibly a unsafe default letting the frontend query whatever they want

bringking:
Hmm maybe, I don't really see it as any less unsafe than say querying for a large list without providing a limit. Is there some damage they could do with a where statement? I guess having a "where" clause that is so tied to the ORM on the front end is kind of a leaky abstraction, but seems easier to pass it directly to sequelize rather than re-implement a certain subset.

mickhansen:
Hmm yeah i suppose you are right. If the user wants to use a magic where argument they should be welcome to.

I've been thinking whether or not we need a more generic arg to where helper that can be utilized and composed in multiple places.

bringking:
So you mean like "where" helper that checks the generic JSON arg for validity?

mickhansen:
Also a set of helpers for converting args into a where statement with a custom mapping function (we have something like that on connections but it's not quite there imo)

chaliy:
I think, `where` implementation should consider to use schema instead of plain JSON. 

P.S. This implementation will not work with GraphQL UI (e.g. 'express-graphql');

bringking:
@chaliy I agree, however I ran into a few issues creating this-

* Modeling the nested $and and $or statements proved to be pretty difficult to model with GraphQL. I can relatively close, then punted to the plain JSON object.

* You end up creating custom "where" types for each queryable model, then you end up with a lot of extra types in the system. 

I think if those issues can be worked around, you are correct. I couldn't get to a resolution. I can spend some more time trying to get $and and $or modeled as input types...

bringking:
Is this going to get merged or are there things I need to change? I think this is a useful stop-gap to give people better query control, while we work on a better implementation

test/unit/replaceWhereOperators.test.js
```diff
@@ -0,0 +1,61 @@
+import chai, {expect} from "chai";
+import {replaceWhereOperators} from "../../src/replaceWhereOperators";
+
+describe("replaceWhereOperators", ()=> {
+  it.only("should take an Object of grapqhl-friendly keys and replace with the correct sequelize operators", ()=> {
```
mickhansen: whoops

mickhansen:
Since this library is still very much experimental i'll merge this when the `.only` is removed.
However i can't promise i won't rip it out again when/if we find a better way to plugin where statements from args.

bringking:
Ah sorry I didn't see that .only. I will remove right away!

bringking:
Okay I fixed that

sebastienbarre:
Hey guys. 
I had no idea this was possible. Great! 
Is this stable and functional in `graphql-sequelize` right now? 

@bringking you were asking "is there some damage they could do with a where statement?" and I can give you a somewhat contrived example that I had to deal with. My application involves `employees`, and the corresponding MySQL table has a `salary` field. Now obviously you do not want to expose everybody's salary (per company policy) so we maintain an access control list, and if an employee does not have the permission to see *other people's* `salary` field, it is never returned by the query. However, if the `where` argument was "free-form" and unconstrained, an employee (or third-party) could still query all employees that have a `salary` between $50,000 and $100,000. The `salary` field would never be returned by the query, but the query would still have used that field as a query condition. Then the third party could query for employees between $50,000 and $75,000, etc., narrowing it down until he/she finds a close enough value for everybody in the company. Just an example :) Technically no "damage" was down to the data, but a problem nonetheless. This doesn't preclude letting people query whatever they want, but at the very least we had to disallow some query conditions based on the fields that were involved and our ACLs...

mickhansen:
@sebastienbarre It's in the project yes, i don't personally use it for the reasons you measoned (hard to control)

PR #165, opened 2016-02-22
Cursor Reconfigure For Order

brad-decker:
### Updates
1. Removed the base64 encoding of order attribute from the cursor. Replaced with the index of the item itself.
2. Updated all the logic based on this to provide true pagination regardless of the order by statement.
3. Added a test case in integration/connections for custom queries and made sure all tests pass there.

### Caveats
1. Need more test cases to make sure nothing can break the pagination. 
2. A unit test for $parent failed on simplifyAST but i haven't touched that. 
Signed-off-by: brad-decker <bhdecker84@gmail.com>

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -128,8 +128,6 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
     });
   }
 
-  let defaultOrderBy = orderByEnum._values[0].value;
-
   before = before || ((options) => options);
 
   let $connectionArgs = {
@@ -143,20 +141,30 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
     return orderBy[0][0];
   };
 
-  let toCursor = function (value, orderBy) {
-    let id = value.get(model.primaryKeyAttribute);
-    let orderValue = value.get(orderByAttribute(orderBy));
-    return base64(PREFIX + id + SEPERATOR + orderValue);
+  /**
+   * Creates a cursor given a item returned from the Database
+   * @param  {Object}   item   sequelize model instance
+   * @param  {Integer}  index  the index of this item within the results, 0 indexed
+   * @return {String}          The Base64 encoded cursor string
+   */
+  let toCursor = function (item, index) {
+    let id = item.get(model.primaryKeyAttribute);
+    return base64(PREFIX + id + SEPERATOR + index);
   };
 
+  /**
+   * Decode a cursor into its component parts
+   * @param  {String} cursor Base64 encoded cursor
+   * @return {Object}        Object containing ID and index
+   */
   let fromCursor = function (cursor) {
     cursor = unbase64(cursor);
     cursor = cursor.substring(PREFIX.length, cursor.length);
-    let [id, orderValue] = cursor.split(SEPERATOR);
+    let [id, index] = cursor.split(SEPERATOR);
 
     return {
       id,
-      orderValue
+      index
     };
   };
 
@@ -171,13 +179,12 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
     return result;
   };
 
-  let resolveEdge = function (item, args = {}, source) {
-    if (!args.orderBy) {
-      args.orderBy = [defaultOrderBy];
-    }
-
+  let resolveEdge = function (item, index, queriedCursor, args = {}, source) {
+    let startIndex = 0;
+    if (queriedCursor) startIndex = Number(queriedCursor.index);
+    if (startIndex !== 0) startIndex++;
     return {
-      cursor: toCursor(item, args.orderBy),
+      cursor: toCursor(item, index + startIndex),
       node: item,
       source: source
     };
@@ -187,11 +194,10 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
     handleConnection: false,
     include: true,
     list: true,
-    before: function (options, args, context, info) {
+    before: function (options, args, context) {
       if (args.first || args.last) {
         options.limit = parseInt(args.first || args.last, 10);
       }
-
       if (!args.orderBy) {
         args.orderBy = [orderByEnum._values[0].value];
       } else if (typeof args.orderBy === 'string') {
@@ -233,43 +239,22 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
 
       if (args.after || args.before) {
         let cursor = fromCursor(args.after || args.before);
-        let orderValue = cursor.orderValue;
+        let startIndex = Number(cursor.index);
 
-        if (model.rawAttributes[orderAttribute].type instanceof model.sequelize.constructor.DATE) {
-          orderValue = new Date(orderValue);
-        }
+        if (startIndex > 0) options.offset = startIndex + 1;
+      }
+      options.attributes = _.uniq(options.attributes);
+      return before(options, args, root, context);
+    },
+    after: function (values, args, root, {source}) {
+      var cursor = null;
 
-        let slicingWhere = {
-          $or: [
-            {
-              [orderAttribute]: {
-                [orderDirection === 'ASC' ? '$gt' : '$lt']: orderValue
-              }
-            },
-            {
-              [orderAttribute]: {
-                $eq: orderValue
-              },
-              [model.primaryKeyAttribute]: {
-                $gt: cursor.id
-              }
-            }
-          ]
-        };
-
-        // TODO, do a proper merge that won't kill another $or
-        _.assign(options.where, slicingWhere);
+      if (args.after || args.before) {
+        cursor = fromCursor(args.after || args.before);
       }
 
-      // apply uniq to the attributes
-      options.attributes = _.uniq(options.attributes);
-
-
-      return before(options, args, context, info);
-    },
-    after: function (values, args, context, {source}) {
-      let edges = values.map((value) => {
-        return resolveEdge(value, args, source);
+      let edges = values.map((value, idx) => {
+        return resolveEdge(value, idx, cursor, args, source);
       });
 
       let firstEdge = edges[0];
@@ -282,6 +267,11 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
       if (model.sequelize.dialect.name === 'postgres' && (args.first || args.last)) {
         if (fullCount === null || fullCount === undefined) throw new Error('No fullcount available');
       }
+      let hasMorePages = false;
+      if (args.first || args.last) {
+        let index = cursor ? Number(cursor.index) : 0;
+        hasMorePages = index + 1 + parseInt(args.first || args.last, 10) < fullCount;
+      }
 
       return {
         source,
@@ -291,8 +281,8 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
         pageInfo: {
           startCursor: firstEdge ? firstEdge.cursor : null,
           endCursor: lastEdge ? lastEdge.cursor : null,
-          hasPreviousPage: args.last !== null && args.last !== undefined ? fullCount > parseInt(args.last, 10) : false,
-          hasNextPage: args.first !== null && args.first !== undefined ? fullCount > parseInt(args.first, 10) : false,
+          hasPreviousPage: hasMorePages,
+          hasNextPage: hasMorePages
         }
       };
     }
--- a/test/integration/relay/connection.test.js
+++ b/test/integration/relay/connection.test.js
@@ -8,6 +8,7 @@ import attributeFields from '../../../src/attributeFields';
 import resolver from '../../../src/resolver';
 import {uniq} from 'lodash';
 
+
 const {
   sequelize,
   Promise
@@ -18,13 +19,10 @@ import {
 } from '../../../src/relay';
 
 import {
-  GraphQLString,
   GraphQLInt,
-  GraphQLFloat,
   GraphQLNonNull,
   GraphQLBoolean,
   GraphQLEnumType,
-  GraphQLList,
   GraphQLObjectType,
   GraphQLSchema,
   graphql
@@ -121,9 +119,19 @@ if (helper.sequelize.dialect.name === 'postgres') {
             values: {
               ID: {value: [this.Task.primaryKeyAttribute, 'ASC']},
               LATEST: {value: ['createdAt', 'DESC']},
+              CUSTOM: {value: ['updatedAt', 'DESC']},
               NAME: {value: ['name', 'ASC']}
             }
           }),
+          before: (options) => {
+            if (options.order[0][0] === 'updatedAt') {
+              options.order = Sequelize.literal(`
+                CASE
+                  WHEN completed = true THEN "createdAt"
+                  ELSE "updatedAt" End ASC`);
+            }
+            return options;
+          },
           connectionFields: () => ({
             totalCount: {
               type: GraphQLInt,
@@ -193,12 +201,12 @@ if (helper.sequelize.dialect.name === 'postgres') {
           orderBy: new GraphQLEnumType({
             name: 'Viewer' + this.Task.name + 'ConnectionOrder',
             values: {
-              ID: {value: [this.Task.primaryKeyAttribute, 'ASC']},
+              ID: {value: [this.Task.primaryKeyAttribute, 'ASC']}
             }
           }),
-          before: (options, args, root) => {
+          before: (options, args, context, {viewer}) => {
             options.where = options.where || {};
-            options.where.userId = root.viewer.get('id');
+            options.where.userId = viewer.get('id');
             return options;
           }
         });
@@ -340,7 +348,7 @@ if (helper.sequelize.dialect.name === 'postgres') {
           name: 'userProject',
           nodeType: this.projectType,
           target: this.User.Projects,
-          before(options){
+          before(options) {
             // compare a uniq set of attributes against what is returned by the sequelizeConnection resolver
             let getUnique = uniq(options.attributes);
             projectConnectionAttributesUnique = getUnique.length === options.attributes.length;
@@ -384,7 +392,7 @@ if (helper.sequelize.dialect.name === 'postgres') {
           })
         });
 
-        let result = await graphql(schema, `
+        await graphql(schema, `
           {
             user(id: ${this.userA.id}) {
               projects {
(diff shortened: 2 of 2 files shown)


Aweary:
:clap:

mickhansen:
Travis isn't currently running the integration tests since a unit test fails (looks like it might be related to a graphql update, not sure).

lib/argsToFindOptions.js
```diff
@@ -0,0 +1,42 @@
...
+          result.order = [[args[key], 'ASC']];
+        }
+      }
+    });
+  }
+
+  return result;
+}
+
+module.exports = exports['default'];
```
mickhansen: lib/ should be in gitignore, either way please remove from PR

src/relay.js
```diff
@@ -212,39 +218,22 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
-        return resolveEdge(value, args, source);
+      var cursor = null;
+
+      if (args.after || args.before) {
+        cursor = fromCursor(args.after || args.before);
+      }
+
+      let edges = values.map((value, idx) => {
+        return resolveEdge(value, idx + 1, cursor, args, source);
       });
```
mickhansen: Would the index not be correct if we're already querying after a cursor? It would be 0 where it should be N + 0

mickhansen:
I've fixed the unit test, please rebase.

src/relay.js
brad-decker: The resolveEdge function takes care of the starting index. Though it could be modified instead of passing the cursor to the function to instead do all the calculations here and pass the resulting index to resolveEdge. Might be more efficient as well.

lib/argsToFindOptions.js
brad-decker: I added the lib folder to this so i could use it straight from github for our application. I can remove it directly before its ready for merge. If you have an alternative suggestion to make this able to be installed from github so we can install versions that aren't ready for NPM for testing etc i'd love to implement that feature.

We use automated build processes on heroku so we are limited to what can be automatically installed.

src/relay.js
mickhansen: Ah right, you're taking the existing cursor and applying based on that.

lib/argsToFindOptions.js
mickhansen: Hm alright, please do remove before we merge.

brad-decker: Will do.

mickhansen:
I generally like the approach of encoding the index in the cursor and then using that index.
The only issue i see is that it does not take into account the connection edges changing (i.e. more being added)

brad-decker:
I'm not familiar with that part of it. Could I alter the mutation test to account for this and add some tests in their for resulting order? Though I do believe that some of that would need to be handled on the client side in the relay mutation configuration using the fields config? Give me some guidance and i can work towards a test case that fails and then work on a resolution.

mickhansen:
I was thinking more of changes to rows from outside graphql-relay.
But i'm not sure that the existing solution solved that any better actually.

test/integration/relay/connection.test.js
```diff
@@ -126,9 +127,19 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
           }),
+          before: (options) => {
+            if (options.order[0][0] === 'updatedAt') {
+              options.order = Sequelize.literal(`
+                CASE
+                  WHEN completed = true THEN "createdAt"
+                  ELSE "updatedAt" End ASC`);
+            }
+            return options;
+          },
```
mickhansen: You can simply add the literal to the `value` of the ENUM i believe.

src/relay.js
```diff
@@ -257,6 +246,11 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
       if (model.sequelize.dialect.name === 'postgres' && (args.first || args.last)) {
         if (fullCount === null || fullCount === undefined) throw new Error('No fullcount available');
       }
+      let hasMorePages = false;
+      if (args.first || args.last) {
+        const index = cursor ? Number(cursor.index) : 0;
```
mickhansen: `let` since it's only used in that scope

src/relay.js
```diff
@@ -212,39 +218,22 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
-      let edges = values.map((value) => {
-        return resolveEdge(value, args, source);
+      var cursor = null;
+
+      if (args.after || args.before) {
+        cursor = fromCursor(args.after || args.before);
+      }
+
+      let edges = values.map((value, idx) => {
+        return resolveEdge(value, idx + 1, cursor, args, source);
```
mickhansen: I prefer we keep the cursor index 0 based, and then do + 1 in the query part.

test/integration/relay/connection.test.js
brad-decker: Didn't work for me because literal returns an object with {val : }

src/relay.js
brad-decker: will make this change today

src/relay.js
brad-decker: Alright will update this today

test/integration/relay/connection.test.js
```diff
@@ -347,6 +358,96 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
+              }
+            }
+          }
+        `);
+        const reordered = correctOrder.data.user.tasks.edges.map(({node}) => {
+          const targetId = fromGlobalId(node.id).id;
+          return this.userA.tasks.find(task => {
+            return task.id === Number(targetId);
+          });
+        });
```
mickhansen: Should we extend this test to also verify that the order from the custom orderBy is correct?

brad-decker: Do you mean that sequelize returns the items in the order we expect from the api? The rest of the test checks the order of the paged queries to match what is returned from relay queries.

mickhansen: The rest of the test tests that paging works against the first result, however we haven't exactly verified the first result :) 
Isn't stricly necessary but does increase robustness.

test/integration/relay/connection.test.js
mickhansen: Hmm, i believe i have that in one of my projects. Having an object shouldn't matter, the `value` in ENUM can be anything AFAIR.

brad-decker:
```
      let orderBy = args.orderBy;
      let orderAttribute = orderByAttribute(orderBy);
      let orderDirection = args.orderBy[0][1];```

specifically this part of the connections configuration prevents that case because its looking directly for an array value. I tried setting it up directly in the order configuration first and it failed.

Comment (mickhansen) [inline test/integration/relay/connection.test.js]: Ah alright, was sure i tried that out, but maybe i used a hook too.

**Comment by mickhansen** [comment]
@brad-decker Any update?

**Comment by mickhansen** [comment]
Do we even need to care about id in the cursor with this update? We could possible just move to the cursor generators provided by graphql-relay

**Comment by chrisbolin** [comment]
hey guys, what's the current status of this?

**Comment by mickhansen** [comment]
@brad-decker has been unavailable for a bit, waiting for him to be back. Do you have a usecase not solved by the current method @chrisbolin?

**Comment by chrisbolin** [comment]
I hit https://github.com/mickhansen/graphql-sequelize/issues/170, which led me here :)

**Comment by mickhansen** [comment]
Ah, i didn't actually envision seeing a lot of MSSQL users i must admit :)

**Comment by chrisbolin** [comment]
Oh, I'm on Postgres. `pageInfo.hasPrevious` is always false though, due to the way the `where` query is created

**Comment by chrisbolin** [comment]
The `where` query looks like this
```
        "id": {
          "$gt": "71"
        }
```
if you have an `after` that evaluates to 71. Therefore `full_count` will not include anything less than 71.

**Comment by chrisbolin** [comment]
`full_count` as: https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L277

**Comment by chrisbolin** [comment]
actually, looking at the code closely, it looks like `hasPreviousPage` will always be false no matter what if you are running a `first` query

**Comment by brad-decker** [comment]
@chrisbolin @mickhansen sorry for the delay... No promises but i should be able to work on this this week.

**Comment by chrisbolin** [comment]
no worries, @brad-decker! Thanks for taking a stab at this.

**Comment by chrisbolin** [comment]
I think this is probably related to the `full_count` issue: The `last` arg seems to generate funky queries. e.g.
```javascript
"where": {
    "$or": [
      {
        "id": {
          "$lt": "17"
        }
      },
      {
        "id": {
          "$gt": "17"
        }
      }
    ]
  },
```

For `example (last:3, before: "YXJyYXljb25uZWN0aW9uJDEwMyQxMDM=")` (where that cursor is decoded to `arrayconnection$103$103`)

mickhansen:
@chrisbolin Hmm yeah, full count is affected by where, odd i haven't encountered this since we do filter by id/order value.

mickhansen:
All in all @brad-decker's update should work better. Not sure why i ruled out index in cursor as a proper fix in the beginning.

brad-decker:
@mickhansen can you do a final review of this and then i'll put our project that is using our fork on pause and remove the lib folder from the repository so you can merge. sorry for such a long delay on this.

mickhansen:
@brad-decker IIRC i was quite happy with most of it, however the diff is near impossible at the moment.
If you could rebase and remove lib to clean it up i'll review asap.

brad-decker:
@mickhansen rebased and removed lib. I've got a failing test that was added in the relay test file regarding model connections that i have to figure out but i'll get on it shortly.

mickhansen:
Still like 30-40 commits in the PR including merge commits.

brad-decker:
Hmm

brad-decker:
Ok successfully rebased, need to do some cleanup tho and fix that test still. Sorry about the false alarm there

brad-decker:
Cleanup done, test passing.

src/relay.js
```diff
@@ -143,20 +141,30 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
   };
 
-  let toCursor = function (value, orderBy) {
-    let id = value.get(model.primaryKeyAttribute);
-    let orderValue = value.get(orderByAttribute(orderBy));
-    return base64(PREFIX + id + SEPERATOR + orderValue);
+  /**
+   * Creates a cursor given a item returned from the Database
+   * @param  {Object}   item   sequelize model instance
+   * @param  {Integer}  index  the index of this item within the results, 1 indexed
```
mickhansen: 0 indexed

src/relay.js
```diff
@@ -171,13 +179,12 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
 
-  let resolveEdge = function (item, args = {}, source) {
-    if (!args.orderBy) {
-      args.orderBy = [defaultOrderBy];
-    }
-
+  let resolveEdge = function (item, index, cursor, args = {}, source) {
+    let startIndex = 0;
+    if (cursor) startIndex = Number(cursor.index);
+    if (startIndex !== 0) startIndex++;
```
mickhansen: I'm not following exactly what goes on here, if we receive a cursor we generate a new cursor?

src/relay.js
```diff
@@ -233,43 +239,23 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
-              },
-              [model.primaryKeyAttribute]: {
-                $gt: cursor.id
-              }
-            }
-          ]
-        };
+        let startIndex = Number(cursor.index);
 
         // TODO, do a proper merge that won't kill another $or
```
mickhansen: Can remove the TODO aswell

src/relay.js
mickhansen: change was not made :)

test/integration/relay/connection.test.js
```diff
@@ -193,12 +201,12 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
-              ID: {value: [this.Task.primaryKeyAttribute, 'ASC']},
+              ID: {value: [this.Task.primaryKeyAttribute, 'ASC']}
             }
           }),
-          before: (options, args, root) => {
+          before: (options, args, context, {viewer}) => {
             options.where = options.where || {};
-            options.where.userId = root.viewer.get('id');
+            options.where.userId = viewer.get('id');
             return options;
```
mickhansen: Should probably be fixed to use context instead of rootValue for viewer.

test/integration/relay/connection.test.js
```diff
@@ -473,7 +481,727 @@ if (helper.sequelize.dialect.name === 'postgres') {
 
         let lastResult = await query(nextResult.data.user.tasks.edges[2].cursor);
         verify(lastResult, lastThree);
-        expect(lastResult.data.user.tasks.pageInfo.hasNextPage).to.equal(false);
+        expect(lastResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
```
mickhansen: Why was this assertion changed?

test/integration/relay/connection.test.js
```diff
@@ -473,7 +481,727 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
+        let nextResult = await query(firstResult.data.user.tasks.pageInfo.endCursor);
+        verify(nextResult, nextThree);
+        expect(nextResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
+
+        let lastResult = await query(nextResult.data.user.tasks.edges[2].cursor);
+        verify(lastResult, lastThree);
+        expect(lastResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
+      });
+
+      it('should support in-query slicing with user provided args/where', async function () {
```
mickhansen: I don't see any args in this test

test/integration/relay/connection.test.js
```diff
@@ -473,7 +481,727 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
+        let nextResult = await query(firstResult.data.user.tasks.pageInfo.endCursor);
+        verify(nextResult, nextThree);
+        expect(nextResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
+
+        let lastResult = await query(nextResult.data.user.tasks.edges[2].cursor);
+        verify(lastResult, lastThree);
+        expect(lastResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
+      });
+
+      it('should support in-query slicing and pagination with first and CUSTOM orderBy', async function () {
```
mickhansen: duplicate test?

test/integration/relay/connection.test.js
```diff
@@ -473,7 +481,727 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
```
mickhansen: duplicate test?

test/integration/relay/connection.test.js
brad-decker: Any suggestions on how to do that? the test was failing when it was just using root, and root.viewer.get('id') saying viewer was not defined. Root would have been context in this case.

test/integration/relay/connection.test.js
brad-decker: my error, confused myself while trying to merge stuff and make tests pass. I'll patch this shortly.

src/relay.js
brad-decker: If there is a cursor supplied to this function, we are simply getting the index of that cursor to determine where to start numbering the next indexes. Because index in this case is going to always start at 0 we need to increment startIndex if we aren't dealing with a startIndex of 0.

Ex: we get the first three items from query, they will have a index of 0,1 and 2 in their cursor returned from the api. If we then send the cursor of the 2 to the after clause, the next 3 returned will start numbering at 2. Without the startIndex++ check the next set of results would be 2, 3 and 4 which would be wrong.

src/relay.js
brad-decker: pretty sure my merging caused this to revert. i've changed it back.

brad-decker:
@mickhansen most of your points have been addressed now. Thanks for the feedback

test/integration/relay/connection.test.js
mickhansen: Usually the test needs to be changed from `graphql(query, {values})` to `graphql(query, null, {values})`

test/integration/relay/connection.test.js
mickhansen: alright, there's a few of them through the tests i believe.

src/relay.js
mickhansen: Ah, of course. Maybe we should rename `cursor` to `queriedCursor`

test/integration/relay/connection.test.js
brad-decker:
```
let result = await graphql(this.schema, `
          {
            viewer {
              tasks {
                edges {
                  cursor
                  node {
                    id
                    name
                  }
                }
              }
            }
          }
        `, null, {
          viewer: viewer
        });
```

Looks like it already follows that format.

mickhansen: Odd, viewer should be on context then.

test/integration/relay/connection.test.js
brad-decker: Yeah all of the ones verifying that we've reached the last page of results should now be reverted back to original state.

test/integration/relay/connection.test.js
brad-decker: Rebase issue i think. I removed

test/integration/relay/connection.test.js
brad-decker: Rebase issue. this was removed.

src/relay.js
brad-decker: I went ahead and renamed this to be clearer. I am holding off on committing until I can make any final changes we may need.

src/relay.js
brad-decker: Fixed

test/integration/relay/connection.test.js
```diff
@@ -476,63 +484,222 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
+        expect(nextResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
+
+        let lastResult = await query(nextResult.data.user.tasks.edges[2].cursor);
+        verify(lastResult, lastThree);
+        expect(lastResult.data.user.tasks.pageInfo.hasNextPage).to.equal(false);
       });
 
-      it('should support nested aliased fields', async function () {
-        let result = await graphql(this.schema, `
+      it('should support in-query slicing with user provided args/where', async function () {
```
mickhansen: Looks like this test was changed although i don't think it was intentional.

test/integration/relay/connection.test.js
```diff
@@ -565,26 +732,26 @@ if (helper.sequelize.dialect.name === 'postgres') {
...
 
         let nextResult = await query(firstResult.data.user.tasks.pageInfo.endCursor);
         verify(nextResult, nextThree);
-        expect(nextResult.data.user.tasks.pageInfo.hasPreviousPage).to.equal(true);
+        expect(nextResult.data.user.tasks.pageInfo.hasNextPage).to.equal(true);
 
         let lastResult = await query(nextResult.data.user.tasks.edges[2].cursor);
         verify(lastResult, lastThree);
-        expect(lastResult.data.user.tasks.pageInfo.hasPreviousPage).to.equal(false);
+        expect(lastResult.data.user.tasks.pageInfo.hasNextPage).to.equal(false);
```
mickhansen: Having a test that does reverse querying is probably still nice to have.

brad-decker:
Thanks @KATT @chrisbolin @mickhansen and everyone else for your patience here. :clap:

KATT:
Amazing!! Thank you so much for all the the time & effort you both put into this @mickhansen + @brad-decker! Can't wait to try it out.

Issue #144, opened 2016-02-04
BigInt columns should be GraphQLFloat


bringking:
Wanted to get your thoughts on this issue we just ran into. 

The graphql spec limits GraphQLInt values to 2^31. 

```
 const num = parseInt(ast.value, 10);
  if (num <= MAX_INT && num >= MIN_INT) {
    return num;
  }
```

However, we are converting BigInt columns to GraphQLInt, which silently returns those over 2^31 as `null`. Changing this type mapping to GraphQLFloat (which feels slightly wrong, since they are not floating point numbers) allows the numbers to be not-null (at least up to the JS limit of 2^53. Thoughts?

mickhansen:
No firm opinion, any clues in the spec?

bringking:
Well the spec specifies that a GraphQLFloat is definitely a FP number- 

> A Float number includes either a decimal point (ex. 1.0) or an exponent (ex. 1e50) or both (ex. 6.0221413e23).

However, the [graphql-js implementation](https://github.com/graphql/graphql-js/blob/master/src/type/scalars.js#L54) just uses `parseFloat`

```js
parseFloat(ast.value)

```
which in our case, passing a large integer value to parseFloat, results in standard integer, e.g.

```js
parseFloat("4121722835") // result: 4121722835
// not
// 4121722835.0
```

So in order to get a BigInt, it seems we have to use either a custom scalar type or use the Float type because of this behavior. I guess the "cleanest" way to do this would be to implement our own Scalar type called BigInt that doesn't adhere to the GraphQLInt limits.  Then the question becomes whether this library starts implementing custom types. In this case, I am leaning towards the custom type as being the best solution...

mickhansen:
A custom scalar is probably the best way to go if there's no standard way of doing int8

brainjam:
The natural mapping would be to GraphQLString wouldn't it?  If you need to further play with it o the GraphQL side you can convert to big-integer or float or whatever.  My use case is that the BIGINT is a 64 bit integer unique iD, so I'm happy enough with the string representation on the GraphQL side.

bringking:
Yeah I agree with @brainjam after talking it over with Lee byron on the other projects. I will do a PR making GraphQLString the default.

PR #167, opened 2016-02-27
Moved BIGINT to default to a GraphQLString 

bringking:
so precision isn't lost, and implemented the rest of the typeMapper unit tests. Fixes issue - https://github.com/mickhansen/graphql-sequelize/issues/144


mickhansen:
Thanks @bringking

PR #169, opened 2016-02-27
Make pageInfo.hasNextPage work for MSSQL

legomind:
Inject full_count column for mssql dialect so that pageInfo.hasNextPage and pageInfo.hasPrevious works for MSSQL


legomind:
Don't merge yet, only works for pageInfo.hasNextPage at the moment.

mickhansen:
Sorry this took a while, actually wanted to merge #165 first, but we're not quite ready to do that yet.

Issue #171, opened 2016-02-28
Custom id resolver


legomind:
`typeResolver ` provides a way to map from a model type to a graphQL type, but as far as I can see there is no way to resolve an id for an entity that is not a sequelize model.

Perhaps something like this?

```javascript
nodeTypeMapper.mapTypes({
  [User.name]: UserType,
  Viewer: ViewerType,
})

nodeTypeMapper.mapIdFetchers({
  Viewer: id => GetViewer(id),
})
```

mickhansen:
Currently it just defaults to the passed type: https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L45

Which is not great in a few cases, had this problem myself that i had to work around.
Supplying your own node resolvers would probably be a good idea, also for models to provide ACL.

mickhansen:
```js
nodeTypeMapper.mapTypes({
  [User.name]: UserType,
  'Viewer': {
    type: ViewerType,
    reslve: (id) => GetViewer(id)
   }
})
```

Might also be an option

legomind:
I like that a lot.

Im willing to work on a PR, if we can nail down a direction.

bringking:
I like the idea of just allowing a function similar to what @mickhansen has-

```
nodeTypeMapper.mapTypes({
  [User.name]: UserType,
  'Viewer': ()=> ViewType
})
```
For flexibility in any mapper. However, in the case of the `idFetcher`, the explicit creation of a another mapping like `mapIdFetchers` makes sense since they are used in different ways and the idFetcher functions will have to be functions that accept the `globalId` as a param.

mickhansen:
@bringking They are actually only ever used for node lookups, so they are sort of tightly related.

bringking:
Right, that makes sense. So based on how it is used in the relay `nodeConnections` function, I think @mickhansen's syntax would be good. In the lib, the `typeResolver` could use the defined `type` field, and the the `idFetcher` could use the resolve function.

mickhansen:
The only problem is that the typeResolver simply gets whatever idFetcher returned and if say idFetcher returns something that's not necessarily a sequelize instance we can't use Model.name.

So in idFetcher we will likely have to append the type somehow (use a prop or whatever).

legomind:
[...]

PR #179, opened 2016-03-06
Add custom id fetcher

legomind:
This PR aims to address #171 
<strike>Not ready to merge just yet, need to add tests, and update docs.</strike>

- [x] Add tests
- [x] Update docs

### DIFF
--- a/docs/relay.md
+++ b/docs/relay.md
@@ -5,6 +5,9 @@
 relay will perform certain queries on a root "node" type.
 graphql-sequelize will automatically map these node lookups to findById calls.
 
+If you wish to use non-sequelize entities, or if you want to override the default
+behaviour for sequelize models, you can specify a resolve function.
+
 ```js
 import {relay: {sequelizeNodeInterface}} from 'graphql-sequelize';
 import sequelize from './your-sequelize-instance';
@@ -31,7 +34,16 @@ const userType = new GraphQLObjectType({
 });
 
 nodeTypeMapper.mapTypes({
-  [User.name]: userType
+  [User.name]: userType,
+
+  //Non-sequelize models can be added as well
+  SomeOther: {
+    type: SomeOtherType, //Specify graphql type to map to
+    resolve(globalId) { //Specify function to get entity from id
+      const { id } = fromGlobalId(globalId);
+      return getSomeOther(id);
+    }
+  }
 });
 
 const schema = new GraphQLSchema({
@@ -154,11 +166,11 @@ const userType = new GraphQLObjectType({
     }
   }
 }
-  
+
 fragment totalCount on userTaskConnection {
    total
 }
-  
+
 fragment getCreated on userTaskEdge {
   wasCreatedByUser
 }
--- a/src/relay.js
+++ b/src/relay.js
@@ -19,42 +19,58 @@ import {
 import _ from 'lodash';
 import simplifyAST from './simplifyAST';
 
-class NodeTypeMapper {
-  constructor(sequelize) {
-    this.models = Object.keys(sequelize.models);
-    this.models.forEach(model => {
-      this[model] = null;
-    });
+export class NodeTypeMapper {
+  constructor() {
+    this.map = { };
   }
 
   mapTypes(types) {
-    Object.keys(types).forEach(type => {
-      this[type] = types[type];
+    Object.entries(types).forEach(([k, v]) => {
+      this.map[k] = v.type
+        ? v
+        : { type: v };
     });
+  }
+
+  item(type) {
+    return this.map[type];
   }
 }
 
 export function idFetcher(sequelize, nodeTypeMapper) {
-  return globalId => {
-    let {type, id} = fromGlobalId(globalId);
-    const models = Object.keys(sequelize.models);
-    if (models.some(model => model === type)) {
-      return sequelize.models[type].findById(id);
+  return async globalId => {
+    const {type, id} = fromGlobalId(globalId);
+
+    const nodeType = nodeTypeMapper.item(type);
+    if (nodeType && typeof nodeType.resolve === 'function') {
+      const res = await Promise.resolve(nodeType.resolve(globalId));
+      res.__graphqlType__ = type;
+      return res;
     }
-    if (nodeTypeMapper[type]) {
-      return nodeTypeMapper[type];
-    }
-    return null;
+
+    const model = Object.keys(sequelize.models).find(model => model === type);
+    return model
+      ? sequelize.models[model].findById(id)
+      : nodeType
+        ? nodeType.type
+        : null;
   };
 }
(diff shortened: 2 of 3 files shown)


docs/relay.md
```diff
@@ -31,7 +34,18 @@ const userType = new GraphQLObjectType({
 });
 
 nodeTypeMapper.mapTypes({
-  [User.name]: userType
+  [User.name]: {
+    type: userType,
+  }
```
mickhansen: The old way should still work

src/relay.js
```diff
@@ -20,41 +20,61 @@ import _ from 'lodash';
...
-    });
+  constructor() {
+    this.map = { };
   }
 
-  mapTypes(types) {
-    Object.keys(types).forEach(type => {
-      this[type] = types[type];
-    });
+  mapTypes( types ) {
```
mickhansen: nitpick but please retain the code style used.

src/relay.js
```diff
@@ -20,41 +20,61 @@ import _ from 'lodash';
...
+    this.map = {
+      ...this.map,
+      ...types,
```
mickhansen: Normalize types from `type` to `{type: type}` here

src/relay.js
```diff
@@ -20,41 +20,61 @@ import _ from 'lodash';
...
+    };
+  }
+
+  item( type ) {
+    const nodeType = this.map[type];
+    return nodeType.type
+      ? nodeType
+      : { type: nodeType };
```
mickhansen: Normalize on set rather than get, then we can just retun this.map[type]

legomind: Ok, I can do that, but my way is more efficient as it avoids iterating the keys of the types object.

docs/relay.md
legomind: Ok, updated the docs so that the new way is only used when necessary.

src/relay.js
legomind: Done.

legomind:
@mickhansen I have updated the PR per your suggestions. The only thing that I haven't done yet is normalize the type map on set. I think my current method of normalizing on get is more efficient.

Do you still want me to normalize the type map on set even though it involves iteration?

legomind:
@mickhansen Ignore my last comment. Babel transpiles the spread operator to iteration anyway.

src/relay.js
legomind: Done.

mickhansen:
Normalizing on get would always be less efficient since it will be called N times for a type where as normalization on set will only ever be called once for a type.

mickhansen:
And ignore my comment i guess since i didn't read your last comment before reading the one before ;)

src/relay.js
```diff
@@ -20,41 +20,59 @@ import _ from 'lodash';
...
+      const res = nodeType.resolve(globalId);
+      res.__graphqlType__ = type;
+      return res;
     }
-    return null;
+
+    const model = Object.keys(sequelize.models).find(model => model === type);
+    return model
+      ? sequelize.models[model].findById(id)
+      : nodeType && nodeType.type
```
mickhansen: No need for the double check with normalization on set.

src/relay.js
```diff
@@ -20,41 +20,59 @@ import _ from 'lodash';
...
-    }
-    if (nodeTypeMapper[type]) {
-      return nodeTypeMapper[type];
+    const {type, id} = fromGlobalId(globalId);
+
+    const nodeType = nodeTypeMapper.item(type);
+    if (nodeType && typeof nodeType.resolve === 'function') {
+      const res = nodeType.resolve(globalId);
+      res.__graphqlType__ = type;
+      return res;
```
mickhansen: We need to add a test that uses an async resolver since i'm not sure this code will work for that. (We need to wait for the promise before adding type)

test/integration/relay.test.js
```diff
@@ -152,12 +160,30 @@ describe('relay', function () {
...
+      [User.name]: { type: userType },
+      [Project.name]: { type: projectType},
+      [Task.name]: { type: taskType },
+      Viewer: { type: viewerType },
+      [customType.name]: {
+        type: customType,
+        resolve(globalId) {
+          const { id } = fromGlobalId(globalId);
+          return generateCustom(id);
+        }
```
mickhansen: Make async to test what i mentioned in previous case. Async is the common case.

legomind: Done.

src/relay.js
legomind: You are absolutely correct. Fixed.

legomind:
@mickhansen The tests pass for me. Not sure why travis failed.

docs/relay.md
```diff
@@ -23,15 +26,22 @@ const userType = new GraphQLObjectType({
   name: User.name,
   fields: {
     id: globalIdField(User.name),
-    name: {
-      type: GraphQLString
-    }
+    name: GraphQLString
```
mickhansen: That's not valid IIRC?

src/relay.js
```diff
@@ -19,42 +19,61 @@ import {
...
-    const models = Object.keys(sequelize.models);
-    if (models.some(model => model === type)) {
-      return sequelize.models[type].findById(id);
+  return async globalId => {
+    const {type, id} = fromGlobalId(globalId);
+
+    const nodeType = nodeTypeMapper.item(type);
+    if (nodeType && typeof nodeType.resolve === 'function') {
+      const resOrPromise = nodeType.resolve(globalId);
+      const res = resOrPromise.then ? await resOrPromise : resOrPromise;
```
mickhansen: It might be cleaner to simply have `await Promise.resolve(nodeType.resolve(globalId))`

src/relay.js
```diff
@@ -19,42 +19,61 @@ import {
...
+    if (!type) {
+      throw new Error(`Unable to determine type of ${ typeof obj }. ` +
+        `Either specify a resolve function in 'NodeTypeMapper' object, or specify '__graphqlType__' property on object.`);
+    }
 
-    return nodeTypeMapper[name];
+    const nodeType = nodeTypeMapper.item(type);
+    return nodeType && nodeType.type
+      ? nodeType.type
+      : null;
```
mickhansen: `nodeType && nodeType.type || null` is cleaner imo, but not overly important.

src/relay.js
legomind: :-) Forgot about that method.

legomind:
@mickhansen Anything else?

mickhansen:
No it's looking good, mind squashing the commits to a few descriptive ones?

legomind:
No problem

legomind:
@mickhansen Done.

mickhansen:
v0.21.0

Issue #172, opened 2016-02-29
Feature proposal: default attributes


jonathanewerner:
Hi,
I have the following problem: For validating permission / access to resources, we need to *always* fetch an owner id like `userId` on the requested instance. This should happen without the need to explicitly require the field in a query.
 
So for example I'd like to be able to say `{ orders(userId: "12375") { id } }`, which should set `findOptions.attributes` to `['userId', 'id']`. This would allow me to a) not overfetch SQL columns and b) to not force the expose unwanted information in the response.

mickhansen:
I had to disable attribute filtering personally because there was always fields i needed that wasn't explicitely requested by the graphql query.

jonathanewerner:
Does that mean that you think that `filterAttributes` = `false` is the way to go here or do you think a resolver option like `defaultAttributes` would make sense?

mickhansen:
That's the work around for now atleast. I like the idea of defaultAttributes though.

jonathanewerner:
I could take a stab at a PR for that if you'd like.

mickhansen:
Sure, please do

jonathanewerner:
I'm trying to get started with dev setup. 
I get test failures when executing `npm run test-docker`:
```
  1) relay connection should support prefetching two nested connections:

      AssertionError: expected 'CAA' to equal 'CCC'
      + expected - actual

      -CAA
      +CCC
      
      at Context.<anonymous> (test/integration/relay/connection.test.js:665:85)
      at [object Generator].next (native)
      at [object Generator].tryCatcher (node_modules/bluebird/js/release/util.js:16:23)
      at PromiseSpawn._promiseFulfilled (node_modules/bluebird/js/release/generators.js:90:49)
      at Promise._settlePromise (node_modules/bluebird/js/release/promise.js:565:26)
      at Promise._settlePromise0 (node_modules/bluebird/js/release/promise.js:605:10)
      at Promise._settlePromises (node_modules/bluebird/js/release/promise.js:684:18)
      at Async._drainQueue (node_modules/bluebird/js/release/async.js:126:16)
      at Async._drainQueues (node_modules/bluebird/js/release/async.js:136:10)
      at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/release/async.js:16:14)

```
This test sometimes fails, sometimes not.
[...]

PR #184, opened 2016-03-12
Enhancement: Resolver option `defaultAttributes`

jonathanewerner:
As discussed in #172, i've implemented the resolver option `defaultAttributes`.

I also fixed the flakyness issue in `resolver.test.js` by not asserting on the order of the compared arrays in two cases.

I wasn't entirely sure about some code style issues that are not answered by the `.eslintrc`:
- `var foo = require('...') vs `import ...`
- `const` vs. `var`?
- arrow fn callbacks vs. `function()`?
- use of es2015/16 features like destructuring -- which ones are "allowed" / which node version is targeted?

It might be a good idea to provide more strict guidance per `.eslintrc`for a more homogenous code look and feel -- I spent the majority of time thinking about those issues and how to "fit in" to the existing code style(s). :)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -88,7 +88,7 @@
     "no-self-compare": 0,
     "no-sequences": 2,
     "no-throw-literal": 2,
-    "no-unused-expressions": 2,
+    "no-unused-expressions": 0,
     "no-void": 2,
     "no-warning-comments": 0,
     "no-with": 2,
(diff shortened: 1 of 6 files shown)


mickhansen:
`import`
`const`
arrow when you need the context, otherwise function
whatever es6 you like, we use babel

you're welcome to update .eslintrc

jonathanewerner:
Ok, great, I'll add some rules to `.eslintrc` then. I'll also create a small `CONTRIBUTING.md` guide with some setup & general advice if you don't mind, just as a starting point. I'll do this in another PR though, this one is finished from my side.

mickhansen:
LGTM, i would appreciate a squash to a few more descriptive commits.

jonathanewerner:
Will do.

jonathanewerner:
I squashed it.

mickhansen:
`v0.22.0`

PR #196, opened 2016-03-31
attributes map type

KATT:


KATT:
I'm aware same result can be achieved by using [`typeMapper.mapType`](https://github.com/mickhansen/graphql-sequelize/blob/master/src/typeMapper.js#L16).

I just think this is better than doing it globally.

mickhansen:
Benefit achieved here is to reuse custom types across a few `attributeFields` calls i assume?

.gitignore
```diff
@@ -8,6 +8,7 @@ pids
 *.seed
 
 .idea
+.DS_Store
```
Glavin001: I recommend creating a Global `.gitignore` file for the future 😉 .
See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #197, opened 2016-03-31
ability to bypass NonNull

KATT:
Useful when using `attributeFields()` in update mutations where some of the fields might be optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,8 @@ attributeFields(Model, {
   exclude: [], // array of model attributes to ignore - default: []
   only: [], // only generate definitions for these model attributes - default: null
   globalId: true, // return an relay global id field - default: false
-  map: {} // rename fields - default: {}
+  map: {}, // rename fields - default: {}
+  allowNull: false, // disable wrapping mandatory fields in `GraphQLNonNull` - default: false
 });
 
 /*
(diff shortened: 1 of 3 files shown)


mickhansen:
`attributeFields` is not originally intended for mutation input, we should probably consider a difference method for to help mutations, one that leaves out readonly attributes by default.

I'll still consider this PR, but lets change the option to `allowNull: true` since that maps to the option in Sequelize.

KATT:
Cool. Do you want me to squash it?

mickhansen:
Yeah that'd be great

bringking:
Can we add a blurb in the docs for this? As a more general point, as we add API options, there should be a blurb of documentation describing the behavior or maybe move to a more formal documentation structure.

mickhansen:
For methods we probably need API documentation.

KATT:
@mickhansen let me know if there's anything else you want me to do with this, or if you decide that you don't want the feature.

README.md
```diff
@@ -162,7 +162,8 @@ attributeFields(Model, {
   exclude: [], // array of model attributes to ignore - default: []
   only: [], // only generate definitions for these model attributes - default: null
   globalId: true, // return an relay global id field - default: false
-  map: {} // rename fields - default: {}
+  map: {}, // rename fields - default: {}
+  allowNull: true, // automatically wrap mandatory fields in `GraphQLNonNull` - default: true
```
mickhansen: The description here is wrong.
Setting allowNull does not wrap mandatory fields, it disabled wrapping, and the default is false.

KATT: amended

PR #218, opened 2016-04-19
Don't deepMerge strings. It causes an infinite loop

idris:
Version 0.21.0 introduced an infinite loop for my app.  In deepMerge, it hits a string and then keeps recursing onto the first character of the string.  This bug was introduced in 77e1edd56cfc205ed24a3ff23916498e78c047b9 when `_.merge` was replaced.

This pull request makes deepMerge only recurse on objects.

I tried to write a test case, but couldn't get it to fail.  This code fixes my app though.  If it's useful, here's the GraphQL in my app that causes the loop:
```
query ViewerQueries {
  viewer {
    _subscriptions2JZtYG: subscriptions(first: 100) {
      edges {
        node {
          _monitors4cE1PA: monitors(first: 1) {
            edges {
              node {
                id
              }
            }
          }
          ...F7
        }
      }
    }
  }
}

fragment F7 on Subscription {
  _monitors4cE1PA: monitors(first: 1) {
    edges {
      node {
        id
      }
    }
  }
}
```

What causes the issue is that I have monitors in a fragment and outside of a fragment.  This is the narrowest case I can make it fail for.  If I remove the fragment F7, it works.

The string that it tries to recurse on is "monitors".

### DIFF
--- a/src/simplifyAST.js
+++ b/src/simplifyAST.js
@@ -2,7 +2,7 @@ function deepMerge(a, b) {
   Object.keys(b).forEach(function (key) {
     if (['fields', 'args'].indexOf(key) !== -1) return;
 
-    if (a[key] && b[key]) {
+    if (a[key] && b[key] && typeof a[key] === 'object' && typeof b[key] === 'object') {
       a[key] = deepMerge(a[key], b[key]);
     } else {
       a[key] = b[key];


src/simplifyAST.js
```diff
@@ -2,7 +2,7 @@ function deepMerge(a, b) {
   Object.keys(b).forEach(function (key) {
     if (['fields', 'args'].indexOf(key) !== -1) return;
 
-    if (a[key] && b[key]) {
+    if (a[key] && b[key] && typeof a === 'object') {
```
mickhansen: Why would `a[key]` be truthy for a string?

mickhansen:
Would love to see a failing unit test for this since i find it very odd.

src/simplifyAST.js
idris: Whoops! It should be `a[key]`.  Will push a fix now

idris:
I'd love to see a failing test as well!  Spent a few hours trying to create one and couldn't get it to fail.  Definitely fails in my app, though.

src/simplifyAST.js
```diff
@@ -2,7 +2,7 @@ function deepMerge(a, b) {
   Object.keys(b).forEach(function (key) {
     if (['fields', 'args'].indexOf(key) !== -1) return;
 
-    if (a[key] && b[key]) {
+    if (a[key] && b[key] && typeof a[key] === 'object') {
```
mickhansen: Probably should be prudent and check b[key] aswell just in case.

idris: I figure if a[key] is not an object, merge should replace it with b[key],
regardless of what b[key] is.

mickhansen: Right, and that happens in the else statement. If it tries to deep merge b[key] as a string into an object it will likely still fail.

idris: Ah, yeah, I see it now. Fixed!

mickhansen:
Great, happy to merge this if you can squash to a single descriptive commit :)

idris:
Squashed.  Thanks for your help!

mickhansen:
v1.1.1

PR #230, opened 2016-05-12
add(resolver): Allow async before and after

janmeier:


src/resolver.js
```diff
@@ -103,12 +103,12 @@ function resolverFactory(target, options) {
     findOptions.context = context;
     findOptions.logging = findOptions.logging || context.logging;
 
-    findOptions = options.before(findOptions, args, context, {
+    findOptions = await Promise.resolve(options.before(findOptions, args, context, {
```
mickhansen: Not using async/await in codebase

mickhansen: Should also be applied here: https://github.com/mickhansen/graphql-sequelize/blob/master/src/generateIncludes.js#L75

test/integration/resolver.test.js
```diff
@@ -1039,6 +1039,59 @@ describe('resolver', function () {
...
+                type: GraphQLInt
+              },
+              order: {
+                type: GraphQLString
+              }
+            },
+            resolve: resolver(User, {
+              before: async function (options) {
+                await Promise.delay(100);
+                return options;
```
mickhansen: returning a promise should be sufficient to test it

janmeier: The resolver expects before to return options

mickhansen: Right, `return Promise.resolve(options)`

janmeier:
Updated - I recommend viewing with `?w=1` since there are a lot of whitespace changes :)

src/generateIncludes.js
```diff
@@ -63,67 +61,70 @@ export default function generateIncludes(simpleAST, type, context, options) {
...
-      }
+      return new Promise(function (resolve) {
+        if (includeResolver.$before) {
+          resolve(includeResolver.$before(includeOptions, args, context, {
+            ast: fieldAST,
+            type: type
+          }));
+        } else {
+          resolve(includeOptions);
+        }
```
mickhansen: Why new promise? If it's because you need a chain `Promise.resolve().then(` might be better.

janmeier: Updated (:

janmeier:
Linted - and added a precommit hook so it doesn't happen again ;)

mickhansen:
1.2.0

PR #244, opened 2016-05-28
[RFC] Support defaultAttributes in sequalizeConnection

### LINKED ISSUE #241: Passing defaultAttributes to relay connection
Is there a way to do this at the moment? Here's what code I'm thinking of:

``` js
export default () => relay.sequelizeConnection({
  name: 'activityComments',
  nodeType: ActivityCommentType,
  target: db.BaseActivity.Comments,
  connectionFields: {
    total: {
      type: GraphQLInt,
      resolve: ({ source }) => source.countComments(),
    },
  },
  defaultAttributes: ['entityId'], // <-- Is there a way of doing this?
});
```

ekosz:
Closes #241

Can't seem to get this working yet. So far I've added a failing test, but my implementation doesn't seem to be working. I thought it could just just collect the option in `sequalizeConnection` and pass it down to the resolver as an extra option, but it seems to be bailing early on `after` before it has time to change the `findOptions`. Any help pointing me to the right place would be super awesome.

I've also done some lite linting on the test file.

### DIFF
--- a/test/integration/relay.test.js
+++ b/test/integration/relay.test.js
@@ -1,15 +1,12 @@
 'use strict';
 
-var chai = require('chai')
-  , expect = chai.expect
-  , resolver = require('../../src/resolver')
-  , helper = require('./helper')
-  , sequelize = helper.sequelize
-  , Sequelize = require('sequelize')
-  , Promise = helper.Promise
-  , sinon = require('sinon')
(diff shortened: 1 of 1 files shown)


mickhansen:
You should be able to implement it in the before hook

PR #245, opened 2016-05-29
Add more documentation on custom connection arguments

### LINKED ISSUE #242: Add documentation on adding custom connectionArgs to connections
I've been reading the source and there [doesn't seem to be a way of modify the connection args](https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L133-L138). There seems to be code that that handles where values being put directly into the connectionArgs and handling them then, but when I try doing that I get a GraphQL error saying: `"Unknown argument \"active\" on field \"goals\" of type \"Company\"."`.

It seems that you must define what where values will be sent up front, but there doesn't seem to be way to do that. There [also seems to be code that would remove them](https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L171-L177) if I did define them.

Is this feature possible currently? It seems to be from the tests, but I just don't know why I would be getting this error but the tests wouldn't be.

ekosz:
Closes #242

### DIFF
--- a/docs/relay.md
+++ b/docs/relay.md
@@ -174,3 +174,40 @@ fragment getCreated on userTaskEdge {
   wasCreatedByUser
 }
 ```
+
+You can pass custom args in your connection definition and they will
+automaticly be turned into where arguments. These can be further modified
+using the `where` option in `sequelizeConnection`.
+
+```js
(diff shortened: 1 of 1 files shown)


mickhansen:
Thanks!

PR #251, opened 2016-06-10
relay: pageInfo: fix hasNextPage and implement hasPreviousPage

LoicMahieu:
- Added new test case `should support pagination with where` for testing pagination when a `where` is applied.
- `hasNextPage` and `hasPreviousPage` now results a correct value


brad-decker:
This is beautiful. :+1:

src/relay.js
```diff
@@ -268,10 +268,21 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
-        hasMorePages = index + 1 + parseInt(args.first || args.last, 10) < fullCount;
+        if (index !== 0) index++;
+
+        hasNextPage = index + 1 + count <= fullCount;
+        hasPreviousPage = index - count >= 0;
+
+        if (args.last) {
+          const _hasNextPage = hasNextPage;
+          hasNextPage = hasPreviousPage;
+          hasPreviousPage = _hasNextPage;
```
mickhansen: `[hasNextPage, hasPreviousPage] = [hasPreviousPage, hasNextPage]`

mickhansen:
Thanks!

Did they not resolve to correct values before? I've been using it in my own projects and seems to be working fine.

The reason they resolve to the same value is that you generally only query with first/after and last/before

LoicMahieu:
Nice value swap ! I'll rebase it!

In fact, I wrote the test `should support pagination with where` when I was working on MySQL integration (PR soon). The values were not correct in that case.
I mostly sure this is correct but feel free to test and approve it ;)

mickhansen:
Thanks @LoicMahieu!

PR #255, opened 2016-06-11
mysql support

LoicMahieu:
In this PR:
- Switch db type with `DB` env variable
- Use `mysql` service in travis
- Add docker-compose file for `mysql`
- Fixes some specific MySQL issues:
  * `mysql: fix limit: should be a integer`: MySQL doesn't like `LIMIT "1"`... `parseInt` solves it.
  * `relay: full_count for mysql`: `OVER()` is not available in MySQL, so we need to insert a sub query.
  * `relay: create count query which respect the where clause`: the sub query need to use the same where clause. With `model.QueryInterface.QueryGenerator.select` we are able to create a correct SQL query. **But Sequelize `scope` is not applied.**
- Added method `beforeRemoveAllTables` to some tests: MySQL does not support multiple foreign keys with the same name, in the same database, even on differents tables... So......... I wrote a small utility which force remove all tables in the database... I think this is a ugly hack but I didn't find any better fix... Ideas are welcome :)

There is a big change in `test/integration/relay/connection.test.js` due to the remove of the `if (helper.sequelize.dialect.name === 'postgres')` wrapping the `describe`.


LoicMahieu:
Tests are randomly failing, same for postgres... I think the order of elements change.

src/relay.js
```diff
@@ -233,6 +233,23 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
           model.sequelize.literal('COUNT(1) OVER()'),
           'full_count'
         ]);
+      } else if (options.limit) {
+        // TODO: scope are not applied to the where clause!
+
+        const countQuery = model
```
LoicMahieu: @mickhansen Do you have idea/tip about the query ?
I looked at [Model definition](https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1280) in Sequelize and there is some methods that transform option before query:
- `this._conformOptions()`
- `this.$injectScope()`
- `this._expandIncludeAll()`
- `Utils.mapFinderOptions()`
- `this._paranoidClause()`

Is there a way to applied it to the option from outside from Sequelize ?
Furthermore is it necessary to clone `options.include` and `options.where` before passing it to `selectQuery` ?

mickhansen: I'd use cloneDeep and conformOptions,
basically you want to do whatever count() does https://github.com/sequelize/sequelize/blob/master/lib/model.js#L1547

LoicMahieu: Nice catch!
The code will look like:
```js
const countOptions = {
  attributes: [['COUNT(*)', 'full_count']],
  include: options.include,
  where: options.where
}

model.$conformOptions(options, model);
model.$injectScope(options);
model.$expandIncludeAll(options);
model.$validateIncludedElements(options);
model._paranoidClause(options);

const countQuery = model
  .QueryInterface
  .QueryGenerator
  .selectQuery(model.tableName, countOptions, model)
  .replace(/;$/, '');
```

I support that methods prefixed with `$` are not "private" but `_paranoidClause` is... Base our implementation on top of private methods is not a good idea...

I tried an other approach: fetch count with `model.count()` but I didn't find a way to pass the result to `after`.

mickhansen: Both `$` and `_` denotes private, in `4.0` well be consistent about that.

mickhansen: You could just fetch the count IN after @LoicMahieu

src/relay.js
```diff
@@ -265,9 +282,6 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
       if (!values[0]) {
         fullCount = 0;
       }
-      if (model.sequelize.dialect.name === 'postgres' && (args.first || args.last)) {
-        if (fullCount === null || fullCount === undefined) throw new Error('No fullcount available');
-      }
```
mickhansen: It should still throw if unavailable

LoicMahieu:
Ok, rebased it with `model.count()` in after when `fullCount` is not found.

Do you think it can be merged as is ?

docker-compose.yml
```diff
@@ -1,16 +0,0 @@
...
-  environment:
-    DB_DATABASE: graphql_sequelize_test
-    DB_USER: graphql_sequelize_test
-    DB_PASSWORD: graphql_sequelize_test
-
-db:
-  image: postgres:9.4
-  environment:
-    POSTGRES_USER: graphql_sequelize_test
-    POSTGRES_PASSWORD: graphql_sequelize_test
```
mickhansen: dont remove compose

src/relay.js
```diff
@@ -265,36 +265,49 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
...
+        // In case of `OVER()` is not available, we need to get the full count from a second query.
+        return model.count({
+          where: argsToWhere(args)
+        })
+          .then(count => {
+            fullCount = count;
+            return result();
+          });
+      } else {
+        return result();
```
mickhansen: async/await is available to simplify this a great deal.

docker-compose.yml
LoicMahieu: I didn't remove it but moved to `test/support` folder. For MySQL, I duplicated the file to be specific to mysql. In package.json, we use the env variable `DB` for switching the `docker-compose` file and the project name.

src/relay.js
LoicMahieu: Huge! Rebased!!

docker-compose.yml
mickhansen: Move it back please, docker-compose and Dockerfile should be at root.
It's possible to have both databases in the same file, it's used for local testing aswell.

test/integration/resolver.test.js
```diff
@@ -18,6 +19,8 @@ import {
 } from 'graphql';
 
 describe('resolver', function () {
+  beforeRemoveAllTables();
```
mickhansen: is sync not sufficient?

LoicMahieu: Unfortunately no... The problem come from the test executed before. It creates tables and `sequelize.modelManager.models = [];` in the before made impossible to reset all tables.

docker-compose.yml
LoicMahieu: Done :)

mickhansen:
Thanks for the great work @LoicMahieu

PR #256, opened 2016-06-13
simplifyAST: resolve array args

LoicMahieu:
In case of an argument is an `GraphQLList`, simplifyAST resolves `undefined` as value. This PR fix it by mapping `arg.values[].value`.


mickhansen:
v2.0.2

PR #258, opened 2016-06-14
relay: mysql: fix fullCount on association

LoicMahieu:
Manual `fullCount` wasn't correct, it didn't care about the relation between the `source` and `target`. Added a new `userB` with some tasks for illustrate the bug.

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -268,14 +268,14 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
 
       if ((args.first || args.last) && (fullCount === null || fullCount === undefined)) {
         // In case of `OVER()` is not available, we need to get the full count from a second query.
+        const options = await Promise.resolve(before({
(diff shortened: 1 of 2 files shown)


src/relay.js
```diff
@@ -268,7 +268,7 @@ export function sequelizeConnection({name, nodeType, target, orderBy: orderByEnu
 
       if ((args.first || args.last) && (fullCount === null || fullCount === undefined)) {
         // In case of `OVER()` is not available, we need to get the full count from a second query.
-        fullCount = await model.count({
+        fullCount = await target.count(source, {
           where: argsToWhere(args)
         });
```
mickhansen: This will likely not work for `target: Association` where association is a N:M association.

LoicMahieu: Complitly right :s

I found `target.manyFromSource.count`... I don't know if it is the good way but it seems to work.

PR #260, opened 2016-06-16
Close #229. Support for commentToDescription option in attributeFields

### LINKED ISSUE #229: attributeFields should transform comment into description
It would be nice if the `attributeFields()` function turned my `comment` on my sequalize models into `description` fields on my graphql schema object types. 

I realize this would not always be desirable, so it could be an option like `comments: true` perhaps? I suppose `false` until you can introduce `true` as a BC.

Apologies if this is already handled somehow and I missed it in the documentation.

Glavin001:
Closes #229.

Usage:
```
Model = sequelize.define(modelName, {
      comment:{
        type: Sequelize.STRING,
        comment: 'This is a comment'
      }
    });
var fields = attributeFields(Model, {
        commentToDescription: true
      });
/* fields = {
  ...
  comment: {
    ...
    description: 'This is a comment'
  }
}
*/
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,8 @@
   "author": "Mick Hansen <maker@mhansen.io>",
   "contributors": [
     {
-      "name": "Glavin Wiechert",
-      "url": "https://github.com/Glavin001"
+      "name": "graphql-sequelize community",
+      "url": "https://github.com/mickhansen/graphql-sequelize/graphs/contributors"
     }
   ],
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,12 @@
     "sequelize"
   ],
   "author": "Mick Hansen <maker@mhansen.io>",
+  "contributors": [
+    {
+      "name": "Glavin Wiechert",
+      "url": "https://github.com/Glavin001"
+    }
+  ],
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


Glavin001:
@mickhansen : I picked `commentToDescription` as per your comment, however I can, of course, easily change that. Let me know what you think is best.

package.json
```diff
@@ -27,6 +27,12 @@
     "sequelize"
   ],
   "author": "Mick Hansen <maker@mhansen.io>",
+  "contributors": [
+    {
+      "name": "Glavin Wiechert",
+      "url": "https://github.com/Glavin001"
```
ofbeaton: It may make more sense when adding a contributors key to make the name `graphql-sequelize community` and the url `https://github.com/mickhansen/graphql-sequelize/graphs/contributors` instead of individually name every person who contributes a pull request.

Glavin001: That works for me, too. @mickhansen any preference?

mickhansen: No preference

Glavin001: Changed it, as @ofbeaton described.

Glavin001:
@mickhansen ready to be reviewed, merged, and published. Thanks!

PR #261, opened 2016-06-16
Close #216. Switch graphql-relay to peerDependency

### LINKED ISSUE #216: switch graphql-relay to peer dependency

Glavin001:
#216

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,11 +33,11 @@
   },
   "homepage": "https://github.com/mickhansen/graphql-sequelize",
   "dependencies": {
-    "graphql-relay": "^0.4.2",
     "invariant": "2.2.1",
     "lodash": "^4.0.0"
   },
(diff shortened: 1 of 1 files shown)


mickhansen:
There was some reason originally why we didn't do this, i can't recall why.

Glavin001:
The reason could be that now users must ensure that have `graphql-relay` already installed so the `peerDependency` is resolved. However, this will be very obvious when users update and find an error stating that `graphql-relay` is missing, and I would assume that they would naturally install it.
That is a problem with not doing it with `peerDependencies` to begin with, that this is almost a breaking change in the sense that it may not already work after they update. However, it is a trivial fix on their end.

brad-decker:
@mickhansen pretty sure the reason was i didn't understand/know about peer dependency at the time.

PR #267, opened 2016-06-21
Support `where` option from arguments, by parsing JSON arguments in s…

JohnyDays:
Where option wasn't actually working in my case, had to make this change to the argument parsing.

Cheers


mickhansen:
Please add a test aswell

JohnyDays:
Added a test and squashed 👍

mickhansen:
v2.1.1

JohnyDays:
Thanks

PR #272, opened 2016-06-30
Fixes #223. Add cache argument to attributeFields

### LINKED ISSUE #223: [Bug] EnumType duplicated
In converting `Sequelize.ENUM` to  `GraphQLEnumType`
When I called twice same model with EnumType, graphql spits schema duplicated.

Glavin001:
Issue #223.

#### Before

```javascript
    // Bad: Will create multiple/duplicate types with same name
    var fields1a = attributeFields(Model);
    var fields2a = attributeFields(Model);
```

#### After

```javascript
    // Good: Will use cache and not create mutliple/duplicate types with same name
    var cache = {};
    var fields1b = attributeFields(Model, {cache: cache});
    var fields2b = attributeFields(Model, {cache: cache});
```

### DIFF
--- a/src/attributeFields.js
+++ b/src/attributeFields.js
@@ -26,13 +26,13 @@ module.exports = function (Model, options = {}) {
 
     if (memo[key].type instanceof GraphQLEnumType ) {
       var typeName = `${Model.name}${key}EnumType`;
-      // Check cache for existing type
+      /*
+      Cache enum types to prevent duplicate type name error
+      when calling attributeFields multiple times on the same model
+      */
       if (cache[typeName]) {
-        // Use cached type
-        // Prevent GraphQL error message
         memo[key].type = cache[typeName];
       } else {
-        // Not found in cache
         memo[key].type.name = typeName;
         cache[typeName] = memo[key].type;
       }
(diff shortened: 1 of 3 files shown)


mickhansen:
Looks like you've based this on an outdated branch or something, please rebase on master.

src/attributeFields.js
```diff
@@ -24,7 +25,18 @@ module.exports = function (Model, options = {}) {
...
     if (memo[key].type instanceof GraphQLEnumType ) {
-      memo[key].type.name = `${Model.name}${key}EnumType`;
+      var typeName = `${Model.name}${key}EnumType`;
+      // Check cache for existing type
+      if (cache[typeName]) {
+        // Use cached type
+        // Prevent GraphQL error message
+        memo[key].type = cache[typeName];
+      } else {
+        // Not found in cache
```
mickhansen: You can remove the multiple comments, and just do one comment ala 'Cache enum types to prevent duplicate type name error when calling attributeFields multiple times on the same model'

Glavin001:
There we go. Branch is all cleaned up 😄 .

Glavin001:
@mickhansen let me know if there are any other changes you think need to be made here before merging. Thanks!

mickhansen:
If you could add a note to the documentation that would be great.

Glavin001:
I have updated the documentation in the README.

PR #279, opened 2016-07-17
Switch bluebird devDependency to dependency

Glavin001:
Resolve `Error: Cannot find module 'bluebird'`


mickhansen:
We're not using bluebird anywhere except for development so should not be necessary.

mickhansen:
Scratch that, async is rewritten to use bluebird of course :)

PR #281, opened 2016-07-18
Fix issue with generateIncludes and __typename

ekosz:
When generateIncludes statements, if there was a `__typename` field present
grahphql-sequelize would try and treat it as a field on the Model and blow up
retrieving its `resolve` function.

This patch instead skips when it encounters a native GraphQL key. We may want
to add more of these later.

This was the error I was getting before:
```
TypeError: Cannot read property 'resolve' of undefined
    at node_modules/graphql-sequelize/lib/generateIncludes.js:37:45
    at Array.map (native)
    at generateIncludes (node_modules/graphql-sequelize/lib/generateIncludes.js:30:52)
    at resolver (node_modules/graphql-sequelize/lib/resolver.js:103:43)
    at resolver node_modules/graphql-sequelize/lib/relay.js:344:14)
```

### DIFF
--- a/src/generateIncludes.js
+++ b/src/generateIncludes.js
@@ -2,6 +2,10 @@ import argsToFindOptions from './argsToFindOptions';
 import {isConnection, nodeAST, nodeType} from './relay';
 import _ from 'lodash';
 
+const GRAPHQL_NATIVE_KEYS = [
+  '__typename',
+];
+
 function inList(list, attribute) {
   return ~list.indexOf(attribute);
 }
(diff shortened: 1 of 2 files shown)


ekosz:
@mickhansen Not sure why the node 6 versions are failing. The error seems unrelated.

mickhansen:
It's unrelated, issue with npm 3 i haven't looked at yet.

src/generateIncludes.js
```diff
@@ -13,6 +13,11 @@ export default function generateIncludes(simpleAST, type, context, options) {
   options = options || {};
 
   return Promise.all(Object.keys(simpleAST.fields).map(function (key) {
+    if (key === '__typename') {
```
mickhansen: could we put the native keys in a constant array instead and match agains that?

ekosz: @mickhansen As far as I can tell from the spec, all other `__`vars start at the root of the document and are for introspection. Only `__typename` is used during queries themselves.

ekosz: But I could move it to an array for future proofing if you'd like

mickhansen: I just like the self-descriptiveness of having a const `GRAPHQL_NATIVE_KEYS`

ekosz:
@mickhansen Updated

mickhansen:
v2.2.1

PR #289, opened 2016-07-25
Allow functions for exclude and only options

ekosz:
This modifies the `exclude` and `only` options of attributeFields to allow
functions to be passed instead of a list of arguments.

This allows more programmatic removal of fields. For example, I would like to
remove all fields that have an underscore in them (foreign keys). This would
allow me to write:

```js
{
  fields: () => ({
    ...attributeFields(Model, { exclude: field => ~field.indexOf('_') })
  })
}
```


mickhansen:
2.3.0

PR #291, opened 2016-07-27
Fixing Pagination Ordering and results

brad-decker:
# The Issue

Paginating with first and last is not easily understandable and results are incorrect if you change direction. 
### Steps to reproduce:
- Query the first 5 items of a connection. Get 5 unique items
- Query the first 5 items of same connection after cursor from the first query. Get 5 new unique items. 
- Query the last 5 items of same connection before cursor from last query. Get 5 new unique items, expected same items as previous query. 
- This is because when you use the last keyword, regardless of your position, the order direction of the query is reversed which only makes sense if you are paginating backwards from the end of the array/results. 
### Steps to fix:
- If we query with the last argument and no before or after cursor, reverse the order direction. When we resolve edges and add their cursors, index starts with the full count of the query and decrements the index instead of incrementing.
- If we are using the last keyword with a before cursor, do not set a offset, instead set a limit equal to the index of the cursor. Then in the after hook reverse the order of the result and trim to match the amount of items asked for with last. 
- No longer necessary to reverse the hasNextPage/hasPreviousPage variables as our index will always be relative to the proper ordering of the results. 
### Improvements
- No longer determine the startIndex in every call to resolveEdges, pass the startIndex into this function instead.

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -282,6 +282,13 @@ export function sequelizeConnection({
         fullCount = 0;
       }
 
+      if ((args.first || args.last) && (fullCount === null || fullCount === undefined)) {
+        // In case of `OVER()` is not available, we need to get the full count from a second query.
+        fullCount = await model.count({
+          where: argsToWhere(args)
+        });
+      }
+
       let startIndex = 0;
       if (cursor) {
         startIndex = Number(cursor.index);
@@ -298,13 +305,6 @@ export function sequelizeConnection({
 
       let firstEdge = edges[0];
       let lastEdge = edges[edges.length - 1];
-
-      if ((args.first || args.last) && (fullCount === null || fullCount === undefined)) {
-        // In case of `OVER()` is not available, we need to get the full count from a second query.
-        fullCount = await model.count({
-          where: argsToWhere(args)
-        });
-      }
 
       let hasNextPage = false;
       let hasPreviousPage = false;


brad-decker:
I'm assuming that there is some difference in the full count returned from postgres vs mysql. Will fix this test soon.

brad-decker:
@mickhansen can we try rerunning the node 4 postgres test?

src/relay.js
```diff
@@ -249,28 +251,31 @@ export function sequelizeConnection({
...
+      let cursor = null;
+      if (args.after || args.before) cursor = fromCursor(args.after || args.before);
 
-        if (startIndex > 0) options.offset = startIndex + 1;
+      if (cursor) {
+        let startIndex = Number(cursor.index);
+        if (startIndex > 0) {
+          options[args.after ? 'offset' : 'limit'] = args.after ? startIndex + 1 : startIndex;
+        }
       }
```
mickhansen: This looks like mostly style changes, ideally we want to keep those out of functional changes so it's easier to review.

mickhansen:
Please add an additional test showing the behaviour you mention in "Steps to reproduce".
Was this an issue with the original implementation that used where statements?

I've been wondering about the refactor to offset/limit generally since it's also slower for higher counts/slices.

src/relay.js
brad-decker: There is functional change baked into the style change. Which portion exactly is looking just like a style change?

brad-decker:
Its an issue where changing direction the middle of pagination isn't supported due to the fact that the indexing changes to match a reverse directional ordering. The previous implementation works ONLY if you paginate from the start to the end, or the end to the beginning but will not work if you change direction in the middle. I'll add a test for this.

src/relay.js
mickhansen: Right you are, not sure why i missed that.

mickhansen:
> If we are using the last keyword with a before cursor, do not set a offset, instead set a limit equal to the index of the cursor. Then in the after hook reverse the order of the result and trim to match the amount of items asked for with last.

So for `last: 5, before: $cursor` you want to fetch in the non-reverse direction?
There's a performance impact of using large OFFSET values.

mickhansen:
@brad-decker Please add a test showing the case from your "steps to reproduce"

test/integration/relay/connection.test.js
```diff
@@ -847,11 +847,9 @@ describe('relay', function () {
...
       expect(sqlSpy.callCount).to.equal(1);
 
-      const nodeNames = result.data.user.projects.edges.map(edge => {
-        return edge.node.tasks.edges.map(edge => {
-          return edge.node.name
-        }).sort()
-      });
+      const nodeNames = result.data.user.projects.edges.map(
+        ({node: {tasks: {edges}}}) => edges.map(({node: {name}}) => name).sort()
+      );
```
mickhansen: Previous code was in my opinion far easier to read. I'm not a huge fan of greatly nested arrow functions

mickhansen:
> Query the last 5 items of same connection before cursor from last query. Get 5 new unique items, expected same items as previous query.

Does this happen in Relay?

mickhansen:
@brad-decker ping

PR #308, opened 2016-08-23
Fix indexOf check for field helper options

Sneagan:
Use the idiomatic method of checking indexOf and by so doing fix false positives on the check.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -181,7 +181,6 @@
     "no-ternary": 0,
     "no-trailing-spaces": 2,
     "no-underscore-dangle": 0,
-    "no-extra-parens": 2,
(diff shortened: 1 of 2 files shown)


mickhansen:
`~` IS idiomatic, if there's a false positive please include a test (it's been working well so far)

Sneagan:
If we must use `~`, I'll submit an alternative solution.

mickhansen:
@Sneagan We musn't necessarily, if you provide a test case where it fails i'm happy to merge your fix :)

Sneagan:
I'll update it with a test, but there is and *existing failing test* that is the result of a tilde in the test. These tildes are out of control.

PR #309, opened 2016-08-24
Fix False Positive While Preserving ~ Operator

Sneagan:
This implementation fixes the error while maintaining the `~` operator. While I'd rather remove this operator for clarity, it is used in both the helper implementation and in the tests for the helper. I was rewriting large portions of the tool and it's probably not my place to reimplement it. I instead added a check for empty arrays on lines 13 and 17 of `attributeFields.js` and added tests to ensure that future empty arrays behave as expected. I also added a warning if both `exclude` and `only` are present. The check is basic but it's better than nothing.

### DIFF
--- a/src/attributeFields.js
+++ b/src/attributeFields.js
@@ -5,13 +5,16 @@ import { globalIdField } from 'graphql-relay';
 module.exports = function (Model, options = {}) {
   var cache = options.cache || {};
   var result = Object.keys(Model.rawAttributes).reduce(function (memo, key) {
+    if (options.exclude && options.only) {
+      console.warn('Filtering by both exclude and only properties is not supported.');
+    }
     if (options.exclude) {
       if (typeof options.exclude === 'function' && options.exclude(key)) return memo;
-      if (Array.isArray(options.exclude) && !~options.exclude.indexOf(key)) return memo;
+      if (Array.isArray(options.exclude) && options.exclude.length !== 0 && ~options.exclude.indexOf(key)) return memo;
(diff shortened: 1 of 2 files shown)


Sneagan:
I also considered the possibility that empty `only` arrays should mean that none of the properties are carried over, but that seems like an unexpected behavior better achieved by placing all fields in the `exclude` array. Further, in my use case, I would like to have both present and empty during development to signal the capabilities of the helper to the rest of my team without having to maintain an `only` array with every property in it.

mickhansen:
In my opinion defining `only: []` should either mean a poorly defined option (why would use attributeFields with `only: []` and throw or it should do what `only: []` "implies" (only nothing).

I'm not much for silent failures on misused options.

Sneagan:
That makes sense,  but it means that passing `only: []` (which is the example in the README) turns the helper into nothing more than a perf problem in disguise, in which case it should at least error. `Error: I am just a waste of cycles when you pass 'only: []'`

mickhansen:
The example in the readme is just meant to convey the possible options and the argument type it takes, not example input - But as mentioned previously i can definitely see how you got there.

> I think only should be ignored unless it has a valid value and [] is not valid for this tool and its goals.

Here's where my dislike of silent failures come in :)
I'd prefer throwing for invalid input

Sneagan:
In that case, would you accept a PR that threw an error for empty `only` arrays?

mickhansen:
Definitely

PR #316, opened 2016-09-15
Remove includes, in favor of dataloader

janmeier:
Almost there, just a couple of intermittent test failures still remain.

Benchmarks can be seen at https://gist.github.com/janmeier/239b764779edc51bad35f547e992a01a

The gist of the gist (...) is this:
* Major speedup for nested hasMany and belongsToMany includes (50-100% faster), both for small and large result sets
* Worse performance for belongsTo. However, even though not as fast as includes, it is still fast (since only a single row is fetched per parent)

Also removed the `filterAttributes` functionality - This can easily be implemented in userland using a `before` hook. The reason for doing this is that if we filter away everything that's not requested, we have to make sure that we still select the foreign keys, which are used to load the next level.
If you wish to do this in your own code, I would recommend only filtering away heavy attributes (big JSON or BLOB columns), not just filtering because you want to cut off a few extra bytes. The graphql idea (as I see it) is that you should only send as little data as requested _over the wire_ - You can still select more data if that's easier.
The AST is available for filtering as `options.ast` (options is the fourth argument) to `before` resolve hooks.


src/resolver.js
```diff
@@ -18,6 +14,8 @@ function validateOptions(options) {
 }
 
 function resolverFactory(target, options) {
+  dataLoaderSequelize(target);
```
mickhansen: Does this automatically map to all the associations in target too?

janmeier: Yep https://github.com/mickhansen/dataloader-sequelize/blob/master/src/index.js#L298

PR #339, opened 2016-10-10
Order By with args

sushantdhiman:
Sometime we require to build `orderBy` query based on `args`, with this PR this will be possible now.

If orderBY ENUM value for any field is a function, then its called with `args` and return value is used to build the query.

```js

function calledWithArgs (args) { return 'name'; };

// Order By ENUM
values: {
   NAME_FUNC: {value: [calledWithArgs, 'ASC']}
}
```

Also this PR will allow proper reversing when order direction is combination of orders like
```sql
DESC NULLS LAST
ASC NULLS LAST
```

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -150,12 +150,9 @@ export function sequelizeConnection({
 
   let orderByDirection = function (orderDirection, args) {
     if (args.last) {
-      orderDirection = orderDirection.indexOf('ASC') >= 0
+      return orderDirection.indexOf('ASC') >= 0
               ? orderDirection.replace('ASC', 'DESC')
               : orderDirection.replace('DESC', 'ASC');
-      orderDirection = orderDirection.indexOf('LAST') >= 0
-              ? orderDirection.replace('LAST', 'FIRST')
-              : orderDirection.replace('FIRST', 'LAST');
     }
     return orderDirection;
   };
--- a/src/relay.js
+++ b/src/relay.js
@@ -144,8 +144,20 @@ export function sequelizeConnection({
     }
   };
 
-  let orderByAttribute = function (orderBy) {
-    return orderBy[0][0];
+  let orderByAttribute = function (orderAttr, args) {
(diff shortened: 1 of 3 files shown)


mickhansen:
To keep things similar the signature for the orderBy callback should probably be `(source, args, context, info)`

src/relay.js
```diff
@@ -144,8 +144,20 @@ export function sequelizeConnection({
...
+  };
+
+  let orderByDirection = function (orderDirection, args) {
+    if (args.last) {
+      orderDirection = orderDirection.indexOf('ASC') >= 0
+              ? orderDirection.replace('ASC', 'DESC')
+              : orderDirection.replace('DESC', 'ASC');
+      orderDirection = orderDirection.indexOf('LAST') >= 0
+              ? orderDirection.replace('LAST', 'FIRST')
+              : orderDirection.replace('FIRST', 'LAST');
```
mickhansen: You're inverting `NULLS FIRST` to `NULL LAST` and vice versa? Not sure that's the best way to go, hmm.

sushantdhiman: I can remove this condition, multi order will still be supported. May be reversing NULLS is not a good idea

mickhansen: Let's remove it for now, better to add later.

test/integration/relay/connection.test.js
```diff
@@ -479,6 +482,84 @@ describe('relay', function () {
...
+            }
+          }
+        }
+      `, null, {});
+
+      expect(this.projectOrderSpy).to.have.been.calledOnce;
+      expect(this.projectOrderSpy.alwaysCalledWithMatch({}, { first: 5 })).to.be.ok;
+    });
+
+    it('should properly reverse NULLS LAST in orderBy', async function () {
```
sushantdhiman: It still test ASC was reversed to DESC, even in multi order case, should keep it

sushantdhiman: But it can be renamed :)

mickhansen:
v3.0.0

PR #357, opened 2016-10-26
ORDER BY multiple columns

manhhailua:
Prefer Sequelize default syntax for better performance.
Expected variables syntax of `order` key:
```
{
  "order": [
    ["createdAt", "ASC"],
    ["updatedAt", "DESC"]
  ]
}
```

### DIFF
--- a/src/argsToFindOptions.js
+++ b/src/argsToFindOptions.js
@@ -19,11 +19,8 @@ export default function argsToFindOptions(args, targetAttributes) {
       }
 
       if (key === 'order' && args[key]) {
-        if (args[key].indexOf('reverse:') === 0) {
-          result.order = [[args[key].substring(8), 'DESC']];
-        } else {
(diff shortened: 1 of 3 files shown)


mickhansen:
Breaks SEMVER.
Not opposed to the idea but you should keep the current input and just wrap in a List (List should support non-list input AFAIR).

iDVB:
+1 Is there a merged solution for this yet?

manhhailua:
@mickhansen : so, you want to keep the current input of "order" as string, don't you? I'm thinking about how to write a List as a String which will be the best for fast parsing.

joeybaker:
Could we do something like `"columnName:desc,columnName,asc"`?

Yehonal:
> Could we do something like `"columnName:desc,columnName,asc"`?

take a look: https://github.com/mickhansen/graphql-sequelize/pull/693

PR #362, opened 2016-11-04
Fixed before call without right arguments

yachaka:
`before` hook was called without args, context or info, and my `before` function depended on it.

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -260,7 +260,7 @@ export function sequelizeConnection({
         let cursor = fromCursor(args.after || args.before);
         let startIndex = Number(cursor.index);
 
-        if (startIndex > 0) options.offset = startIndex + 1;
+        if (startIndex >= 0) options.offset = startIndex + 1;
       }
       options.attributes = _.uniq(options.attributes);
       return before(options, args, context, info);
--- a/src/relay.js
+++ b/src/relay.js
@@ -295,9 +295,15 @@ export function sequelizeConnection({
         }, args, context, info));
 
         if (target.count) {
-          fullCount = await target.count(options);
+          // If it's a relation
+          if (target.associationType) {
+            fullCount = await target.count(source, options);
+          } else {
+            // target: Model case
+            fullCount = await target.count(options);
+          }
         } else {
-          fullCount = await target.manyFromSource.count(options);
+          fullCount = await target.manyFromSource.count(source, options);
         }
       }
 
--- a/src/relay.js
+++ b/src/relay.js
@@ -295,9 +295,9 @@ export function sequelizeConnection({
         }, args, context, info));
 
         if (target.count) {
-          fullCount = await target.count(source, options);
+          fullCount = await target.count(options);
         } else {
-          fullCount = await target.manyFromSource.count(source, options);
+          fullCount = await target.manyFromSource.count(options);
         }
       }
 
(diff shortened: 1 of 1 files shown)


src/relay.js
```diff
@@ -288,12 +292,12 @@ export function sequelizeConnection({
...
           where: argsToWhere(args)
-        }));
+        }, args, context, info));
 
         if (target.count) {
-          fullCount = await target.count(source, options);
+          fullCount = await target.count(options);
         } else {
-          fullCount = await target.manyFromSource.count(source, options);
+          fullCount = await target.manyFromSource.count(options);
```
mickhansen: These lines should not be changed

yachaka: Are you talking about this one line 300 ? Or the ones above also ?
Because without theses fixes, the connection doesn't work. But maybe the change line 300 was unnecessary.

mickhansen: Both lines to counting should remain unchanged.
Master is green previously so it does work.

/cc @janmeier

janmeier: Not sure about this one actually. Target can both be an association (in which case passing the source instance as the first arg is correct (https://github.com/sequelize/sequelize/blob/master/lib/associations/has-many.js#L342)), but it can also be a model - In which case we shouldn't pass the source.

Probably missing a count test for the latter case

mickhansen: True, there's likely an issue with `target: Model`

yachaka: Yes, that's the one I'm talking about. My case is a `target: Model`, and so in this case, I don't think we should pass the `source` argument. I can try to make my changes work with previous tests, and add some for `target: Model`.

mickhansen: @yachaka target in both cases can also be an association, which is the common case, which you break :)

yachaka: @mickhansen Sure. I'm talking about fixing this while providing support for `target: Model`, which is my case. Do you want me to fix that ?

mickhansen: @yachaka That would be great :)

yachaka: @mickhansen I'm looking at this tonight ! ;)

yachaka:
@mickhansen @janmeier Added check for `associationType`, and fixed additional edge case when `after` cursor index is 0. Please review and comment ;)

src/relay.js
```diff
@@ -196,9 +196,14 @@ export function sequelizeConnection({
   };
 
   let resolveEdge = function (item, index, queriedCursor, args = {}, source) {
-    let startIndex = 0;
+    let startIndex = false;
```
mickhansen: `null` is likely better

src/relay.js
```diff
@@ -288,10 +297,16 @@ export function sequelizeConnection({
...
-        }));
+        }, args, context, info));
 
         if (target.count) {
-          fullCount = await target.count(source, options);
+          // If it's a relation
+          if (target.associationType) {
+            fullCount = await target.count(source, options);
+          } else {
+            // target: Model case
```
mickhansen: comments are probably not needed here :)

src/relay.js
```diff
@@ -301,8 +316,12 @@ export function sequelizeConnection({
       let hasPreviousPage = false;
       if (args.first || args.last) {
         const count = parseInt(args.first || args.last, 10);
-        let index = cursor ? Number(cursor.index) : 0;
-        if (index !== 0) index++;
+        let index = cursor ? Number(cursor.index) : false;
```
mickhansen: `null`

mickhansen:
v4.0.2

PR #363, opened 2016-11-07
In line with documentation

yachaka:
In the documentation, it is specified to use `Model.name` in `nodeTypeMapper`, but it actually takes the type from `Model.options.name.singular`, which led me to an error when trying to query the `node` field.


mickhansen:
We should probably support looking for both.

yachaka:
@mickhansen I think `.name` is always set on the Model. So, supporting only the `.name` isn't a bad thing. Supporting `.options.name.singular` is deceptive, as you expect it to match your model name, and not the singular version.

PR #380, opened 2016-11-29
renamed fieldASTs to fieldNodes

### LINKED ISSUE #369: sequelizeConnection breaks with latest version of graphql
sequelizeConnection breaks with latest version of graphql (0.8) with error

```
Cannot read property '0' of undefined
```. 

Most probably due to these breaking changes:
https://github.com/graphql/graphql-js/releases/tag/v0.8.0

dervos:
fixes #369 

see https://github.com/graphql/graphql-js/releases/tag/v0.8.0

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -345,7 +345,7 @@ export function sequelizeConnection({
   });
 
   let resolver = (source, args, context, info) => {
-    if (simplifyAST(info.fieldASTs[0], info).fields.edges) {
+    if (simplifyAST(info.fieldNodes[0], info).fields.edges) {
       return $resolver(source, args, context, info);
     }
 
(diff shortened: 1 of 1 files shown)


legomind:
This change needs to be backwards compatible with older versions of graphql-js

mickhansen:
@dervos please change the code so it supports looking for both keys.

mickhansen:
You might want to target master with this fix instead as graphql-relay does not appear to be too happy with the new graphql

dervos:
@mickhansen does this workaround satisfy the need for backwards compatibility?

The build fails on this branch but succeeds on [mine](https://travis-ci.org/dervos/graphql-sequelize/jobs/179709073), perhaps you could re-run it?

mickhansen:
@dervos It does, but you'll need to open against master as there are dependency conflicts.

PR #401, opened 2017-01-10
remove redundant eslint rule breaking build

jakemmarsh:
I kept the newest definition, which sets `no-extra-parens` to 2.


mickhansen:
Thanks @jakemmarsh

PR #421, opened 2017-02-01
Update graphql to ^0.9.0 and add temporary name for EnumType in type …

klimentLambevski:
…mapper

### DIFF
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -86,6 +86,7 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
 
   if (sequelizeType instanceof ENUM) {
     return new GraphQLEnumType({
+      name: 'tempEnumName', //grpahql 0.9 requires name for EnumType. Using tempEnumName until its replaced with field generated name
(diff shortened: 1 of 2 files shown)


src/typeMapper.js
```diff
@@ -86,6 +86,7 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
 
   if (sequelizeType instanceof ENUM) {
     return new GraphQLEnumType({
+      name: 'tempEnumName', //grpahql 0.9 requires name for EnumType. Using tempEnumName until its replaced with field generated name
```
mickhansen: Will break if you have multiple enums in your schema, which would be most.

klimentLambevski: the name is replaced after type mapper is called so there is no problem for now, it is not the best solution  but maybe you can suggest better

toxsick:
Hi guys,

is there any progress on this? Would like to upgrade to graphql ^0.9.0 too.

regards
Hannes

mickhansen:
PR has lint errors

PR #422, opened 2017-02-02
argsToWhere where(key, value) -> where(key, value, previousWhere)

jakesiebers:
The where(key, value) is run for each custom arg to generate the proper
where statement for that arg.

The issue is that if two arguments interact with the same field, the
second one will overwrite the first when it is assigned to the previous
where.
e.g. You have a sampleFieldGreaterThan arg and a sampleFieldLessThan
arg. You might fill out both if you are doing a range. The second one
would overwrite the first.

The simple fix: pass the previous where values to the function so you
can perform merging logic or any other complex logic inside the where.


mickhansen:
Can you add a test aswell please?

jakesiebers:
Added test :)

Issue #426, opened 2017-02-12
relay: after hook misses context and info


majodev:
Hi!

See [here](https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L332). I think the outer `.after` handler should also receive the respective `context` and `info` arguments, possibly with the typical resolve function signature: `(result, args, context, info) => result`.

Cheers.

mickhansen:
Yeah, that makes sense.

majodev:
Perfect. Pull --> https://github.com/mickhansen/graphql-sequelize/pull/427

PR #427, opened 2017-02-13
relay after hook: function signature

majodev:
* pass args, context and info to match general resolve function signature

See https://github.com/mickhansen/graphql-sequelize/issues/426


mickhansen:
v5.2.0

PR #438, opened 2017-02-24
Handle an array of order options for dynamic filters

tobyhede:
This change will simply pass through order options to sequalize, allowing for multiple columns to be handled. 

```
{
  order: {
    type: new GraphQLList( new GraphQLList(GraphQLString) ),
    description: 'A JSON object conforming the the shape specified in http://docs.sequelizejs.com/en/latest/docs/querying/#ordering'
  }
}
```

### DIFF
--- a/src/argsToFindOptions.js
+++ b/src/argsToFindOptions.js
@@ -19,7 +19,9 @@ export default function argsToFindOptions(args, targetAttributes) {
       }
 
       if (key === 'order' && args[key]) {
-        if (args[key].indexOf('reverse:') === 0) {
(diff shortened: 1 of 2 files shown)


mickhansen:
I'm generally opposed to adding generic magic, i think graphql schemas should be more explicit than generic.

Introduce this as a helper of sorts rather than a default thing.

tobyhede:
Makes sense. My use case is a grid component that allows for filtering and sorting, so the explicit path ends up being quite complicated as all permutations would need to be covered. 

Do you have an idea for the best way to make it an opt in?

Issue #417, opened 2017-01-30
MySQL full_count not propagated


toxsick:
Hi everybody,

we are using graphql-sequelize with MySQL. A while ago #255 added full_count support for relay connections with MySQL. We are using connection arguments to implement search functionality. The full_count variable represents the total number of results matching the search. This is crutial information we want to display in the frontend. However I think there is currently no way to access this information outside of the relay connection implementation here:

https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L303-L310

Is there a way to access the full_count for MySQL, without out performing another extra count query?

regards Hannes

mickhansen:
Currently not, but no reason we couldn't add it to: https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L332

toxsick:
@mickhansen yes that is what I thought. So we should add a `fullCount` to the object passed to the after hook?

I can create a pr for that.

regards Hannes

mickhansen:
@toxsick that should be sufficient yeah

PR #441, opened 2017-03-08
add relay connection fullCount data to after hook

toxsick:
Hi mickhansen,

as we discussed in #417 here is my shot on propagating fullCount info to the after hook of a relay connection. I tried to extend the testcase too, but I'm not sure if it is the right place.

I don't think any of the existing docs have to be updated for this, because the data passed to the after hook doesn't seem to be documented.

Sorry that it took me a while, but you know how it is :)

Feel free to give me feedback.

regards
Hannes


mickhansen:
v5.3.0

PR #465, opened 2017-04-14
Upgrade peer dependencies

rpellerin:
`peerDependencies` need to be upgraded because they cause warning in projects since newer versions of `graphql` and `graphql-relay` were recently releases.


rpellerin:
Any updates on this?

PR #466, opened 2017-04-20
Add sequelize through attributes to the edge resolver

percevalw:
Allows to resolve edge fields to through model attributes
```
{
    users {
        username
        friends {
            edges {
                createdAt // <--- allow to add this ( cf the sequelize model definition )
                node {
                    username
                }
            }
        },
    }
}
```

By adding it and its type to the sequelizeConnection options :
```
const UserFriendsConnection = sequelizeConnection({
  name: 'UserFriends',
  nodeType: UserType,
  target: User.Friends,
  edgeFields: {
    createdAt: { type: GraphQLString },
  },
});
```

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -18,6 +18,7 @@ import {
 
 import _ from 'lodash';
 import simplifyAST from './simplifyAST';
+import assert from 'assert';
 
 export class NodeTypeMapper {
   constructor() {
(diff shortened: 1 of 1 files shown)


src/relay.js
```diff
@@ -204,10 +205,29 @@ export function sequelizeConnection({
       startIndex = 0;
     }
 
+    assert(
+      !item.$options.includeNames || item.$options.includeNames.length === 1,
+      'Edge must match to exactly one through-model'
+    );
```
mickhansen: Needs to be removed, resolveEdge is also exposed as a helper.

PR #502, opened 2017-06-29
fix crash if enum starts with a digit or has two consequent illegal symbols

Alexsey:
Examples of enum values that crashes now are `'22'` and `'foo--bar'`. This will fix it

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,9 +381,16 @@ attributeFields(Model, {
 
 ### ENUM attributes with non-alphanumeric characters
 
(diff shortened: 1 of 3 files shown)


Alexsey:
I have made it with two commits - first is a fix and second is a small refactoring. If you don't like the refactoring - just tell me and I will remove second commit

Alexsey:
And I don't quite understand what is about crash on travis. If this is something I should handle - please advise me how to fix it

PR #504, opened 2017-07-04
pass context for non-edge fields

sushantdhiman:


src/relay.js
```diff
@@ -356,7 +356,8 @@ export function sequelizeConnection({
     return {
       source,
       args,
-      where: argsToWhere(args)
+      where: argsToWhere(args),
+      context
```
mickhansen: This shouldn't be required, graphql always passed context to all resolvers

sushantdhiman: Changes in context are not observed for non edge fields, I think it keep using old context from higher levels somehow. I will try to add a test to demonstrate this.

mickhansen: Context generally shouldn't change

PR #505, opened 2017-07-11
Resolver options docs

wbobeirne:
Document the second argument of `resolver`, the options object. Also provides an example of one by creating a user "search". I might have been a little too vague about `handleConnection`, and `before` and `after`'s `context` args. I also left out `info`, since I couldn't find that in GraphQL's own docs.

This PR should address issues #4 & #5.


mickhansen:
Thanks a ton @wbobeirne

Issue #468, opened 2017-04-24
defaultListArgs does not support query variables (eg. variable value for `where` becomes null)


pre:
Query variables with array values do not seem to work. The same query works when the array values are passed directly in the query. However, if the array values are passed using query variables, the resolver just sees `null`. 

The following query loses the variable value of `$nicks` which becomes `null` in the resolver:

```graphql
query Users($nicks:[String!]) {
  users(where: {nick: {
    	in: $nicks} }) {
    id
    lastname 
    firstname    
  }
}

variables:
{"nicks": ["Carrie", "Tim"]}
```

The query works, when I pass the values directly in the query without using variables: 
```graphql
query Users($nicks:[String!]) {
  users(where: {nick: {
    	in: ['This', 'Works']} }) {
    id
    lastname 
    firstname    
  }
}
```

The route looks like this:
```javascript
users: {
  type: new GraphQLList(userType),
  args: defaultListArgs(),
  resolve: resolver(models.User, {
    before: (opts, args) => {
      console.log("!!!!!!!!!! before opts where", opts.where.nick);
      console.log("!!!!!!!!!! before args", args);
      return options;
    }
  })
}
```

And the debugging output looks like:
```
!!!!!!!!!! before opts { where: { nick: { '$in': null } }, 
  attributes: [ ...attrs here... ], logging: undefined }
!!!!!!!!!! before args { where: { nick: { in: null } } }
```

It doesn't matter whether I pass the array as `in: [$nicks]` or `in: $nicks`, the resulting value is always null (either `null` or `[Object]` where `Object` is `null` when looked into). 

I have spent quite some time trying to figure out this, so any tips about the direction where to debug further or fix this, would be welcome.

pre:
Possibly related issue: https://github.com/mickhansen/graphql-sequelize/issues/250

mickhansen:
graphql-sequelize doesn't modify args, do you have the same issue with pure graphql?

pre:
You are right. I also get null without using graphql-sequelize:

```javascript
        resolve: (root, args) => {
          console.log("!!!!!!!!!! root", root);
          console.log("!!!!!!!!!! args", args);
        }
```

outputs 
```
!!!!!!!!!! root undefined
!!!!!!!!!! args { where: { flowdockNick: { in: null } } }
```

I am using graphql-server-express, would you suggest this is a bug in that project?

pre:
[...]

PR #506, opened 2017-07-18
Improve where

Alexsey:
This pull request addresses two issues:
1. It was impossible to pass "where" as query variable. Now supported in form of JSON string
2. It was impossible to pass parmeters inside "where" as query variables https://github.com/mickhansen/graphql-sequelize/issues/468. Now it can be done

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,24 @@ defaultArgs(Model);
 */
 ```
 
+If you would like to pass "where" as a query variable - you should pass it as a JSON string and declear it's type as SequelizeJSON:
+
+```
+/* with GraphiQL */
+// request
+query($where: SequelizeJSON) {
+  user(where: $where) {
+    name
+  }
+}
+
+// qyery variables
+# because you can't use single quotes you need to escape JSON's the double quotes
+{
+  "where": "{\"name\": {\"like\": \"Henry%\"}}"
+}
+```
+
 ### defaultListArgs
 
 `defaultListArgs` will return an object like:
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -1,9 +1,20 @@
 import { GraphQLList } from 'graphql';
+import _ from 'lodash';
 import argsToFindOptions from './argsToFindOptions';
 import { isConnection, handleConnection, nodeType } from './relay';
 import invariant from 'assert';
 import Promise from 'bluebird';
 import dataLoaderSequelize from 'dataloader-sequelize';
+
+function whereQueryVarsToValues(o, vals) {
+  _.forEach(o, (v, k) => {
+    if (typeof v === 'function') {
+      o[k] = o[k](vals);
+    } else if (v && typeof v === 'object') {
+      whereQueryVarsToValues(v, vals);
+    }
+  });
+}
 
 function resolverFactory(target, options) {
   dataLoaderSequelize(target);
@@ -47,6 +58,11 @@ function resolverFactory(target, options) {
     findOptions.logging = findOptions.logging || context.logging;
 
     return Promise.resolve(options.before(findOptions, args, context, info)).then(function (findOptions) {
+      if (args.where && !_.isEmpty(info.variableValues)) {
+        whereQueryVarsToValues(args.where, info.variableValues);
+        whereQueryVarsToValues(findOptions.where, info.variableValues);
+      }
+
       if (list && !findOptions.order) {
         findOptions.order = [[model.primaryKeyAttribute, 'ASC']];
       }
--- a/src/types/jsonType.js
+++ b/src/types/jsonType.js
@@ -5,6 +5,7 @@ import {
   GraphQLBoolean,
   GraphQLString
 } from 'graphql';
+import _ from 'lodash';
 
 import { Kind } from 'graphql/language';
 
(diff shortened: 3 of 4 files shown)


mickhansen:
I'll take a solid look at this but i'm generally learning towards less query magic as i definitely feel graphql should be more explicit in arguments accepted - graphql-sequelize will likely move towards being purely resolver helpers (i think attributeFields is slightly harmfull in the long term also, atleast it should be refactored so that `fields` must be defined)

janmeier:
Adding official support makes users think that this is the correct way to do it (which I think it generally isn't). It would open users up to a plethora of data leaks.

In my opinion, this is similar to accepting SQL in a query parameter and using it without any sanitation

Alexsey:
@janmeier but it doesn't add anything that was not there before: where already could accept any values. I have just add support of query variables for it

mickhansen:
Fair point @Alexsey, discussions regarding removing this functionality will be for the next major

mickhansen:
v5.4.0

RichAyotte:
> In my opinion, this is similar to accepting SQL in a query parameter and using it without any sanitation

Ouch, I hope not. I thought Sequelize sanitized the parameters. Can you show an example injection using `where` in this way?

janmeier:
Not in the "traditional" sense of injecting `DROP TABLE users`, but in the sense of leaking more data than the user might think.

Alice whats to allow Bob to look for users matching either idA or idB, so she exposes the following interface:
```graphql
query($where: SequelizeJSON) {
  user(where: $where) {
    name
  }
}
```

Which Bob uses
```graphql
user({ where: { id: { $or: [idA, idB] }}})
``` 

However, Eve comes along
```graphql
user({ where: { id: { $gt: 0 }}})
```

And suddenly she loaded all users in the database, even though the endpoint only ever legimately needed to be used to load two users.

This is not a direct leak on information (hopefully Alice wrote some more checks to ensure that Bob and Eve are only allowed to load users they can actually access).

Explicitly adding arguments (in this case `ids`, perhaps with a limit of how many you can load) forces you to think about what your frontend actually needs to query on, and taking the corresponding precautions.

This is also slightly similar to 
```js
router.get('/user', () => {
  return User.findAll({
    where: {
      id: req.query.id
    }
  });
});

// /user?id=42 Gives one user
// /user?id={ $gte: 0 } - Gives all users, if json parsing is enabled for request params
```

RichAyotte:
Access control is definitely more difficult with this kind of flexibility.

To guarantee that the permissions for each table have been defined and are applied to each query, this should probably be done.

```js
new Sequelize(
    ...
    , {
        define: {
            hooks: {
                beforeFind
            }
        }
    }
)

function beforeFind(options) {
    if (!options) {
        throw new Error(`Options missing from query.`)
    }

    if (!options.user) {
        throw new Error(`User has insufficient privileges.`)
    }

    if (!this.permissions) {
        throw new Error(`Permissions for ${this.name} are not defined. Access Denied.`)
    }
    
    // Get the user role and reduce query scope if required.
    const scopedOptions = optionsToScopedOptions({
        permission: scopeOptionToPermission(options.scope)
        , options
    })
    return scopedOptions
}

```

janmeier:
That seems like a good pattern for solving this pattern generally, and making sure it's applied to all queries

RichAyotte:
One last thing. I think that this type of guarding should be documented.

I ran into a out of memory issue today.

```text
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory

<--- Last few GCs --->

[31179:0x4c8226f590]   271959 ms: Mark-sweep 1410.2 (1475.5) -> 1410.1 (1459.5) MB, 5952.4 / 0.0 ms  (+ 0.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 5953 ms) last resort 
[31179:0x4c8226f590]   277766 ms: Mark-sweep 1410.1 (1459.5) -> 1410.1 (1459.5) MB, 5806.6 / 0.0 ms  last resort 

<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x160fc859bbd9 <JS Object>
    1: _then [/home/rich/Projects/omn/backend/node_modules/bluebird/js/release/promise.js:~219] [pc=0x384d43e742ff](this=0x11a9b8b68669 <a Promise with map 0x32ce334e49c1>,didFulfill=0x1a66609d08a1 <JS Function (SharedFunctionInfo 0x2a56be6fde1)>,didReject=0x160fc8582241 <undefined>,_=0x160fc8582241 <undefined>,receiver=0x160fc8582241 <undefined>,internalData=0x160fc8582241 <undefined>)
```

Had a bunch of queries with 40,000 and 50,000 digits in the IN() clause.
```sql
SELECT *
FROM `ProjectServiceAddresses` AS `ProjectServiceAddresses`
WHERE `ProjectServiceAddresses`.`project_service_id`
IN (101, 102 ... (40,000 INTs) ... 55338, 55339)
ORDER BY `ProjectServiceAddresses`.`id` ASC;
```
I ended up adding this guard in the beforeFind hook.

```js
// Enforce a limit of 1000 rows and log when it's exceeded.
if (!options.limit || options.limit > 1000) {
  options.limit = 1000
  log.warn(`SQL limit clause missing or exceepds ${options.limit}. Forcing to ${options.limit}.`)
  log.debug(options)
}
```
It's  the same defense strategy as the malicious where clause.

PR #511, opened 2017-07-30
fix 5.4.0 regression of 'where' as parameter

Alexsey:
https://github.com/mickhansen/graphql-sequelize/issues/508

Issue #471, opened 2017-04-27
Extending the default args to allow includes


tnrich:
Hey there,

Thanks for your awesome repo. I've been using it so far via `graphql-sequelize-crud` and am liking how it sets up most things I need automagically. It appears to use your `defaultListArgs` function under the hood. 

One thing I can't figure out how to do with it is to make a query that makes a sequelize "includes" call on an associated model. 

 For example, if I want to query my Projects for a given search string. I can use the where argument to search the todo fields. 
```
Project.findAll({
    where: {  // where is supported out of the box
        "status": {
            "like": "%bla%"
         },
    }
})
```
But, searching on the associated owner name doesn't seem possible:

```
Project.findAll({
    include: [{  //include is not supported
        model: User,
        where: {
            "status": {
                "like": "%bla%"
             },
        }
    }]
})
```

Do you know how I can support this type of call to sequelize? Should I just try to extend the default args to accept an "include" type as well?

Any guidance/help would be greatly appreciated. Thanks!

mickhansen:
I'm not sure i follow, you want the client to be able to define what includes to make? That should be known implicitely from the requested fields.

tnrich:
Hey @mickhansen, thanks for getting back to me. Let me try to explain more clearly:
Here is the query I'm making to filter a list of materials (searching all possible fields for the string "a"):
```
query ($limit: Int, $order: String, $where: SequelizeJSON, $offset: Int) {
  materials(limit: $limit, order: $order, where: $where, offset: $offset) {
    id
    status
    sourceSystem
    type
    name
    createdAt
    updatedAt
    user {
      id
      fullName
      __typename
    }
    __typename
  }
}
```
with variables:
```
{
  "limit": 15,
  "order": "",
  "where": {
    "$or": {
      "status": {
        "like": "%a%"
      },
      "sourceSystem": {
        "like": "%a%"
      },
      "type": {
        "like": "%a%"
      },
      "name": {
        "like": "%a%"
      },
      
    }
  },
  "offset": 0
}
```

Now, all that is working great, but, I also want to filter on the user.fullName. I'm not sure how I can do this using graphql-sequelize/graph-ql-sequelize crud. I can't pass variables to my nested "user" query. Nor does the top-level query "where:" support filtering on nested data (at least I don't think). 

So, is there a way to filter on the user.fullName? 

Thank you very much for your help! Keep up the good work!

Alexsey:
[...]

PR #516, opened 2017-08-10
Upstream include

Alexsey:
There are 4 commits
Second commit is appending https://github.com/mickhansen/graphql-sequelize/pull/511 - it ads regression tests and docs
Last commit is addressing https://github.com/mickhansen/graphql-sequelize/issues/471

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,87 +414,264 @@ fullName: {
 
 ### defaultArgs
 
-`defaultArgs(Model)` will return an object containing an arg with a key and type matching your models primary key and
-the "where" argument for passing complex query operations described [here](http://docs.sequelizejs.com/en/latest/docs/querying/)
+`defaultArgs(Model)` will return an object containing:
+ * an arg with a key and type matching your models primary key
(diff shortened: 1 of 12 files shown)


mickhansen:
This feels like an anti pattern to me, i'll look at getting this merged.
But in the next major version i'll look to remove `where` completely as i think a core library should not encourage such behaviour, we can move it to a userland plugin.

Alexsey:
Just add [required: true](https://github.com/mickhansen/graphql-sequelize/pull/516/files#diff-aaf2fbbae86572f96182d7e0646276e4R17) for all includes

Alexsey:
@mickhansen As far as I'm familiar with this part of the code base I would be happy to move `where` and `include` in to separate plugins if an API to "plug" would be provided :)

Alexsey:
I did rebase on master and now builds are falling. Doesn't looks like tests - more like something of infrastructure. If it requires any changes from me - please let me know because I don't know how could I affect the build results

PR #520, opened 2017-09-08
Make v2.4.* compatible with graphql >= 0.9.0 (#510)

maksim-nw:
Copied from 3b4fbb31b064b84eb66b2451475a5d872c4de9c8


mickhansen:
That commit is already in master though?

Issue #524, opened 2017-10-06
$between: [Date1, Date2] does not work correcty in association's before hook


matthiasklan:
I have the following strange behavior, when I try to filter out associations according to an date range. The first time I run the query everything works fine. But when I change my range and run the query after the first one, it will use the old range. I logged the findOptions and the new data is present. If I run only the second query I get the expected result, but again not after multiple fetches. With other fields like IDs it works without any problems. 

```javascript
    const query = `
        query Q( $branchId: ID!, $range: [Date], $userIds: [ID] ) {
          fetchBranch(id: $branchId) {
            id,
            shifts(range: $range, userIds: $userIds) {
              id            
            }
          }
        }
      `;
```
```javascript
//resolver
Branch: {
    shifts: resolver(models.Branch.associations.Shifts, {
      before: (findOptions, args) => {
          findOptions.where = args.where || {};
          // works only for the first query correctly
          if(args.range) findOptions.where.start = { $between: args.range };
          // works anytime
          if(args.userIds) findOptions.where.userId = { $between: args.userIds };
          return findOptions;
        },
      }
    )
```

If I query directly the Shifts (not the Branch association) and use the identical before hook, everything works as expected.

Does it has something to do how dates are parsed and evaluated? thank you

mickhansen:
What do you mean by first or second query? Do you mean first and second execution of the same query?
What does the SQL look like?

matthiasklan:
Hi, yes I mean the first and second execution of the same query. In my test I simply execute the same query, but with different range parameters. In my scenario I have a calendar and want to get data for the next or previous month. Here is the log of my findOptions and the SQL queries. Note the exact same range in the second SQL query.

```
findOptions of first query execution:

"where": {
        "start": {
          "$between": [
            "2017-10-06T05:16:21.078Z",
            "2017-10-07T11:16:21.000Z"
          ]
        }
      }


 SELECT "id", "start" FROM "Shifts" AS "Shift" WHERE ("Shift"."shifts" IN (1) AND "Shift"."start" BETWEEN '2017-10-06 05:16:21.078 +00:00' AND '2017-10-07 11:16:21.000 +00:00') ORDER BY "Shift"."id" ASC;

```

```
findOptions of second query execution (identical query but different range value:

"where": {
        "start": {
          "$between": [
            "2017-10-06T05:16:21.104Z",
            "2017-10-06T10:16:21.104Z"
          ]
        }
      }

[...]

PR #526, opened 2017-10-07
$between: [Date1, Date2] does not work correcty in association's before hook

matthiasklan:
if changing parameters of the range, the new parameters won't get integrated into the SQL
(see issue [524#](https://github.com/mickhansen/graphql-sequelize/issues/524#issuecomment-334933013)). To run the tests use the test:between script


package.json
```diff
@@ -17,6 +17,7 @@
     "build:docker": "docker-compose build",
     "test:docker": "DIALECT=${DIALECT:=postgres} docker-compose run dev /bin/sh -c \"npm run test:integration\"",
     "test:integration": "mocha $npm_package_options_mocha test/integration/*.test.js test/integration/**/*.test.js",
+    "test:between": "mocha $npm_package_options_mocha test/integration/between.test.js",
```
mickhansen: not needed, mocha supports `.only` etc

mickhansen:
@matthiasklan Please see if you can make the code much smaller, all the code can't possible be required to show the defect

mickhansen:
```
{
  "createdAt": {
    "$between": [
      "2014-06-10T00:00:00.000Z",
      "2014-06-15T00:00:00.000Z"
    ]
  }
}
{
  "createdAt": {
    "$between": [
      "2014-06-10T00:00:00.000Z",
      "2014-06-18T00:00:00.000Z"
    ]
  }
}
```

Where statements look fine to me, your test is probably incorrect.

test/integration/between.test.js
```diff
@@ -0,0 +1,424 @@
...
+                projectId: this.projectB.id
+              }
+            ]
+          }, {
+            include: [User.Tasks]
+          })
+        ).bind(this).spread(function (userA, userB) {
+          this.userA = userA;
+          this.userB = userB;
+          this.users = [userA, userB];
```
mickhansen: Not using userB, why create it?

mickhansen:
```
{
  "createdAt": {
    "$between": [
      "2014-06-10T00:00:00.000Z",
      "2014-06-18T00:00:00.000Z"
    ]
  }
}
Executing (default): SELECT `id`, `title`, `created_at` AS `createdAt`, `userId`, `projectId` FROM `tasks` AS `task` WHERE (`task`.`userId` IN (1) AND `task`.`created_at` BETWEEN '2014-06-10 00:00:00.000 +00:00' AND '2014-06-15 00:00:00.000 +00:00') ORDER BY `task`.`id` ASC;
```
SQL is incorrect though, odd.

test/integration/between.test.js
matthiasklan: I will remove all the unneeded code.

mickhansen:
@matthiasklan Please cut the code way down, going through so many lines of irrelevant code is annoying and a waste of time.

mickhansen:
The issue is resolved if direct input are used instead of graphql variables.

matthiasklan:
Ok I will try to replace GraphQLDate with GraphQLString in my project. thank you very much for your help. The bug is really odd, because it works as expected, if I use the exact same before hook directly on Tasks filtered by createdAt and the userID.

mickhansen:
The bug must be related to how caching is done in dataloader-sequelize and somehow date objects/GraphQLDate break it

mickhansen:
Fixed in dataloader-sequelize@1.5.3

PR #527, opened 2017-10-07
Sequelize where error

julesGoullee:
Sequelize version  [4](http://docs.sequelizejs.com/manual/tutorial/upgrade-to-v4.html) refuse  usage string as operator,
Btw, when use where with `SequelizeJSON` got :

> Invalid value [object Object]

By usage of symbole fix this error, but I need help for one test I skipped I'm not understand why It will passed, it's not a SequelizeJSON, pur json not escape

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "homepage": "https://github.com/mickhansen/graphql-sequelize",
   "dependencies": {
     "dataloader-sequelize": "^1.5.0",
+    "sequelize": "^4.13.5",
     "invariant": "2.2.1",
     "bluebird": "^3.4.0",
     "lodash": "^4.0.0"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -70,7 +71,6 @@
     "mysql": "^2.11.1",
     "pg": "^5.0.0",
     "pg-hstore": "^2.3.2",
-    "sequelize": "^3.24.6",
```
mickhansen: Sequelize 3 still needs to be supported

src/replaceWhereOperators.js
```diff
@@ -37,25 +40,25 @@ function replaceKeyDeep(obj, keyMap) {
  */
 export function replaceWhereOperators(where) {
   return replaceKeyDeep(where, {
```
mickhansen: retain old code, have two maps available depending on the availability of `Op`

mickhansen:
> Sequelize version 4 refuse usage string as operator,

No it doesn't? It just does a warning IIRC?

Issue #151, opened 2016-02-12
ENUM characters ¼ ½ ¾


RichAyotte:
An empty value is returned when splitting on ¼ ½ ¾ and an error is thrown on https://github.com/mickhansen/graphql-sequelize/blob/master/src/typeMapper.js#L95

``` javascript
const specialChars = /[^a-z\d_]/i;
const value = '¾'

sanitizedValue = value
.split(specialChars)
.reduce((reduced, val, idx) => {
    let newVal = val;
    if (idx > 0) {
        console.log(idx, val)
        newVal = `${val[0].toUpperCase()}${val.slice(1)}`;
    }
    return `${reduced}${newVal}`;
})
```

https://jsfiddle.net/RichAyotte/x25qm45r/

mickhansen:
That's a specific case

Alexsey:
This merge request https://github.com/mickhansen/graphql-sequelize/pull/502 should fix it

RichAyotte:
#502 didn't solve this issue and I'm not sure what the right approach is but here's what I propose. Since GraphQL requires the key to be a string that matches `/^[_a-zA-Z][_a-zA-Z0-9]*$/ ` We could transform the [¼](http://www.fileformat.info/info/unicode/char/00bc/index.htm) [½](http://www.fileformat.info/info/unicode/char/00BD/index.htm) [¾](http://www.fileformat.info/info/unicode/char/BE/index.htm) characters to their HTML entity names.
```
¼ - > 'frac14'
½ -> 'frac12`
¾ -> 'frac34'
```
If this is acceptable, would you prefer a PR for just these characters or should others be included?

mickhansen:
@RichAyotte I'm quite amazed people have enums with those values. But yes some sort of map is probably the best way to go, as long as it's consistent and tested

RichAyotte:
Fixed in https://github.com/mickhansen/graphql-sequelize/pull/531

PR #531, opened 2017-10-19
ENUM special characters translation map support

RichAyotte:
This patch adds support for special characters such as ¼ in ENUM data types. These special characters are translated to their HTML entity names.

### DIFF
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -56,10 +56,12 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
     JSON
   } = sequelizeTypes;
 
-
-
-  // Regex for finding special characters
-  const specialChars = /[^a-z\d_]/i;
+  // Map of special characters
+  const specialCharsMap = new Map([
+    ['¼', 'frac14']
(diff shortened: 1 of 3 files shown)


RichAyotte:
Fixes #151

src/typeMapper.js
```diff
@@ -56,10 +56,12 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
...
   } = sequelizeTypes;
 
-
-
-  // Regex for finding special characters
-  const specialChars = /[^a-z\d_]/i;
+  // Map of special characters
+  const specialCharsMap = new Map([
+    ['¼', 'frac14']
+    , ['½', 'frac12']
```
mickhansen: rest of the codebase uses trailing commas

mickhansen:
Would love a few tests on the generated enum name, rather than that one test that just verifies it doesnt break

RichAyotte:
@mickhansen styling fixed and tests added. Thanks.

mickhansen:
v5.4.4

Issue #534, opened 2017-11-03
Mutations + DataLoader causes issues


idris:
I'm trying to upgrade `graphql-sequelize` to the latest version (using DataLoaders).  I have a problem, especially in tests (but can happen in normal cases), where the DataLoader is cached, so the result of the Mutation doesn't show up in future queries.

For example:
1. do a GraphQL query, which triggers something that gets cached in a DataLoader
2. do a mutation that affects some data in that query's results
3. in the output of the mutation, the same DataLoader from step 1 (and step 2) is used, thereby returning stale data.  The data was updated in step 2, but the DataLoader still has it cached.

I think we need to call `resetCache` on `dataloader-sequelize` each time a new Query/Mutation is issued.  But then we also need to call `resetCache` after any mutation (before it returns any output).

mickhansen:
https://github.com/mickhansen/dataloader-sequelize/issues/39#issuecomment-341648109

idris:
#536 would also resolve this issue

bkoltai:
So do I understand correctly that the suggested resolution to stale versions of objects being returned from the cache after an update is to disable the `dataloader-sequelize` always in the `resolver` `options`? Or is this in conjunction with some fixes to `dataloader-sequelize`?

I'm having this same issue but want to continue to benefit from `dataload-sequelize` for `Queries`.

Thanks for these great libraries, btw!

PR #535, opened 2017-11-04
fieldASTs -> fieldNodes

idris:
This commit changed `info.fieldASTs` to `info.fieldNodes`: https://github.com/graphql/graphql-js/commit/e49df4980396ed40d9b43bad738c53e001d8f0e5#diff-15f6b6c92ff954f80995d31726f89851

In this branch, I've made `graphql-sequelize` v2.4 support both.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "graphql-sequelize",
-  "version": "2.4.0",
+  "version": "2.4.1",
   "description": "GraphQL & Relay for MySQL & Postgres via Sequelize",
   "main": "lib/index.js",
   "options": {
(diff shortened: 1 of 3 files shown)


idris:
(this will need a version bump to 2.4.2 as well)

mickhansen:
I published a patch on that old version, but that's a one-off, not looking to maintain multipe majors :)

idris:
Thanks!  I'd love to upgrade the major version but I cannot because of this issue: https://github.com/mickhansen/graphql-sequelize/issues/534

mickhansen:
A contribution for https://github.com/mickhansen/graphql-sequelize/issues/536 would be most welcome - Should be pretty easy too :)

mickhansen:
But going forward, when dataloader implements per-request batching i'd love to just remove dataloader from here and just direct people to it in the readme.

PR #546, opened 2017-11-24
Issue/545 graphqlid for uuid

galki:


src/attributeFields.js
```diff
@@ -45,6 +51,15 @@ module.exports = function (Model, options = {}) {
 
     }
 
+    // set integer and string primaryKey attriubte types to GraphQLID
+    if ( attribute.primaryKey === true
+      && ( memo[key].type === GraphQLInt
+        || memo[key].type === GraphQLString
```
mickhansen: type doesn't really matter i suppose

mickhansen: not sure if this should be behind an option, also remember to add a test

galki: option: good idea

"remember"? you mean learn from scratch lol

PR #548, opened 2017-11-24
Resolver: introduce `associationInclude` option

### LINKED ISSUE #547: Feature request: disable use of include in resolver
In a collection resolver, I need to deal with complex includes. The resulted query does not provide correct values for the nested associated objects.

At this time, the `resolver()` will [use the nested association if it is defined](https://github.com/mickhansen/graphql-sequelize/blob/master/src/resolver.js#L72-L80). I would like to disable this behavior by option so it will always fallback to the association getter.

I propose to create a new option `associationInclude` (idea for better name?) (default: true).

My corrent workarround, remove object from parent values before going to the resolve 😔
```js
partner: {
  type: types.partnerType,
  resolve: (() => {
    const resolve = resolver(Litter.Partner)
    return async (source, ...args) => {
      source.dataValues.partner = undefined
      return resolve(source, ...args)
    }
  })()
}
```

LoicMahieu:
Fix #547

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -34,6 +34,7 @@ function resolverFactory(target, options = {}) {
   if (options.before === undefined) options.before = (options) => options;
   if (options.after === undefined) options.after = (result) => result;
   if (options.handleConnection === undefined) options.handleConnection = true;
+  if (options.associationInclude === undefined) options.associationInclude = true;
 
   resolver = function (source, args, context, info) {
     var type = info.returnType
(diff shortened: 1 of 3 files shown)


src/resolver.js
```diff
@@ -69,7 +70,7 @@ function resolverFactory(target, options = {}) {
       }
 
       if (association) {
-        if (source.get(association.as) !== undefined) {
+        if (options.associationInclude && source.get(association.as) !== undefined) {
```
mickhansen: In this case the data has already been fetched, wrong place to exclude it.

mickhansen:
Includes are already disabled in favor of dataloader

Issue #539, opened 2017-11-09
BIGINT in Sequelize v4 gets translated to GraphQLInt


idris:
It should be `GraphQLString`.

This code in Sequelize makes `BIGINT` inherit from `INTEGER`:
https://github.com/sequelize/sequelize/blob/0047dca9dffe9133ae231d7e0de50bca5714634a/lib/data-types.js#L213

Which causes `graphql-sequelize` to short-circuit and return `GraphQLInt` instead of `GraphQLString`: https://github.com/mickhansen/graphql-sequelize/blob/9bb7dd36596faf4e5b4707c21ba2b92176048581/src/typeMapper.js#L71-L73

This test fails if run with Sequelize v4:
https://github.com/mickhansen/graphql-sequelize/blob/9bb7dd36596faf4e5b4707c21ba2b92176048581/test/unit/typeMapper.test.js#L45-L49

mickhansen:
Agreed, PR welcome, will publish as major

RichAyotte:
@idris Should be fixed in latest version.

PR #551, opened 2017-11-30
Upgrade Sequelize to v4, making sure BIGINT gets mapped to GraphQLString

stevehollaar:
Fixes https://github.com/mickhansen/graphql-sequelize/issues/539

If possible, can this also be cherry-picked to a `2.x` release?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "graphql-sequelize",
-  "version": "9.1.0",
+  "version": "9.2.0",
   "description": "GraphQL & Relay for MySQL & Postgres via Sequelize",
   "main": "lib/index.js",
   "options": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -70,7 +70,7 @@
     "mysql": "^2.11.1",
     "pg": "^5.0.0",
     "pg-hstore": "^2.3.2",
-    "sequelize": "^3.24.6",
+    "sequelize": "^4.23.2",
```
mickhansen: need to test against ^3 aswell

RichAyotte: I've tested this patch with ^3 and ^4 and it works. Adding a dev dependencies for both versions is currently not possible with npm (https://github.com/npm/npm/issues/5499) However, it is with yarn. Would you consider using yarn?

RichAyotte: Also, would you accept the patch with the changes in package.json discarded for now?

mickhansen: > Would you consider using yarn?

Definitely consider yarn, use it a lot privately.

> Also, would you accept the patch with the changes in package.json discarded for now?

yes

PR #559, opened 2018-01-22
Migrated infra to support node >= 6

amilajack:


package-lock.json
```diff
@@ -0,0 +1,5897 @@
+{
```
mickhansen: No reason to have a lock file i believe?

amilajack:
Also there aren't any tests for this repo. Can you manually test these changes locally?

mickhansen:
What do you mean there aren't any tests?
There are both tests and travis :)

amilajack:
Ahh never mind. I was confusing this project with another one

.babelrc
```diff
@@ -1,7 +1,13 @@
...
-  "plugins": ["transform-object-rest-spread"],
+  "plugins": [
+    "transform-object-rest-spread"
+  ],
   "presets": [
-    "async-to-bluebird",
-    "es2015-node4"
+    ["env", {
+      "targets": {
+        "node": 6
```
mickhansen, line 8: think we could probably go 8 and then drop async to bluebird

PR #568, opened 2018-02-21
support resolver/relay target as thunk

mickhansen:


src/resolver.js
```diff
@@ -15,31 +15,32 @@ function whereQueryVarsToValues(o, vals) {
...
   invariant(options.include === undefined, 'Include support has been removed in favor of dataloader batching');
   if (options.before === undefined) options.before = (options) => options;
   if (options.after === undefined) options.after = (result) => result;
   if (options.handleConnection === undefined) options.handleConnection = true;
 
-  resolver = function (source, args, context, info) {
-    var type = info.returnType
-      , list = options.list || type instanceof GraphQLList
+  return async function (source, args, context, info) {
+    let target = typeof targetResolver === 'function' && targetResolver.findAndCountAll === undefined ?
```
janmeier: Resolver might be a bit confusing word here, I think thunk would be better

src/relay.js
```diff
@@ -189,6 +181,8 @@ export function sequelizeConnection({
   let argsToWhere = function (args) {
     let result = {};
 
+    if (where === undefined) return result;
+
     _.each(args, (value, key) => {
       if (key in $connectionArgs) return;
```
intellix: args always contains orderBy but now it's not necessarily inside $connectionArgs so orderBy can end up inside the WHERE clause: #570
Should we force a check here for key equaling `orderBy` or is there a fix to be applied elsewhere?

mickhansen: We currently set `args.orderBy`, that should be rewritten, it's not exactly side effect free or obvious unless you look at the source

PR #571, opened 2018-02-24
Update Relay toCursor() to work with Sequelize 4+

### LINKED ISSUE #569: Support for Sequelize >=4
upgrade to version 7.0.0, this version don't support Sequelize >=4.

` let toCursor = function toCursor(item, index) {

    let id = item.get(item.Model.primaryKeyAttribute);

    return (0, _base.base64)(PREFIX + id + SEPERATOR + index);
  };
`
it seem to change 

` let toCursor = function toCursor(item, index) {

        let id = item.get(targetMaybeThunk.primaryKeyAttribute);

        return (0, _base.base64)(PREFIX + id + SEPERATOR + index);
  };
`

intellix:
@mickhansen @flybluewolf

Fixes #569

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -158,7 +158,7 @@ export function sequelizeConnection({
    * @return {String}          The Base64 encoded cursor string
    */
   let toCursor = function (item, index) {
-    let id = item.get(item.Model.primaryKeyAttribute);
+    let id = item.get(item.constructor ? item.constructor.primaryKeyAttribute : item.Model.primaryKeyAttribute);
     return base64(PREFIX + id + SEPERATOR + index);
   };
 


src/relay.js
```diff
@@ -158,7 +158,7 @@ export function sequelizeConnection({
    * @return {String}          The Base64 encoded cursor string
    */
   let toCursor = function (item, index) {
-    let id = item.get(item.Model.primaryKeyAttribute);
+    let id = item.get(item.constructor.primaryKeyAttribute);
```
mickhansen: breaks v3 support, we need both

intellix: should be fixed, I've never touched v3 :)

PR #577, opened 2018-03-22
Add fieldConfig property to sequelizeConnection return value

jedwards1211:
Now instead of wasting keystrokes with
```js
    tasks: {
      type: userTaskConnection.connectionType,
      args: userTaskConnection.connectionArgs,
      resolve: userTaskConnection.resolve
    }
```
You can just write
```js
  tasks: userTaskConnection.fieldConfig
```

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -366,6 +366,11 @@ export function sequelizeConnection({
     nodeType,
     resolveEdge,
     connectionArgs: $connectionArgs,
-    resolve: resolver
+    resolve: resolver,
+    fieldConfig: {
+      type: connectionType,
+      args: $connectionArgs,
+      resolve: resolver
+    }
   };
 }


jedwards1211:
I think the timeouts in the CI tests are too low?

docs/relay.md
```diff
@@ -142,11 +142,7 @@ const userType = new GraphQLObjectType({
     name: {
       type: GraphQLString
     },
-    tasks: {
-      type: userTaskConnection.connectionType,
-      args: userTaskConnection.connectionArgs,
-      resolve: userTaskConnection.resolve
-    }
+    tasks: userTaskConnection.field
```
mickhansen: add something that still shows the structure of the value

mickhansen: i almost never use the resolver as-is, as i suspect is the case for many, so we still want to highlight how that's done.

jedwards1211: > i almost never use the resolver as-is
It doesn't seem like there are any examples in the documentation of using the resolver in custom ways.  I had assumed that if there were a good reason to *not* use the resolver as-is then `before` and `after` resolve hooks would be pointless.  But I can show the structure of the value

jedwards1211:
@mickhansen does it look good now?

docs/relay.md
```diff
@@ -201,14 +201,14 @@ const userType = new GraphQLObjectType({
...
+      type: userTaskConnection.field.type,
       args: {
-        ...userTaskConnection.connectionArgs, // <-- Load the defaults
+        ...userTaskConnection.field.args, // <-- Load the defaults
         titleStartsWith: { // <-- Extend further yourself
           type: GraphQLString,
         }
       },
-      resolve: userTaskConnection.resolve
+      resolve: userTaskConnection.field.resolve,
```
mickhansen: retain the old structure

docs/relay.md
```diff
@@ -143,9 +143,9 @@ const userType = new GraphQLObjectType({
       type: GraphQLString
     },
     tasks: {
-      type: userTaskConnection.connectionType,
-      args: userTaskConnection.connectionArgs,
-      resolve: userTaskConnection.resolve
+      type: userTaskConnection.field.type,
+      args: userTaskConnection.field.args,
+      resolve: userTaskConnection.field.resolve,
```
mickhansen: i meant more something like `/* {type, args, resolve} * /`

jedwards1211: as a comment next to `tasks: userTaskConnection.field` you mean?

mickhansen: yeah just document the structure, leave as much documentation as is otherwise and just add an extra example

docs/relay.md
```diff
@@ -142,11 +142,7 @@ const userType = new GraphQLObjectType({
     name: {
       type: GraphQLString
     },
-    tasks: {
-      type: userTaskConnection.connectionType,
-      args: userTaskConnection.connectionArgs,
-      resolve: userTaskConnection.resolve
-    }
+    tasks: userTaskConnection.field, /* {type, args, resolve} * /
```
mickhansen: i just noticed that the old style is no longer documented, which it absolutely should be, so please restore the original documentation and then add additions as new pieces of documentation.

jedwards1211: You know, I've found in my own work that this PR isn't really as useful as I had first imagined.  maybe it will just add confusion.  Do you think I should finish up the documentation or should we just close this?

mickhansen: Fine to close by me.

PR #579, opened 2018-03-23
support the alias name of field for order by

flybluewolf:
support the alias name of field for order by in relay


mickhansen:
Please follow the general code style.
Please add a test covering this addition.

In addition to that, what's the general motivation? Sequelize already handles this.

PR #595, opened 2018-05-09
Treat non nullable lists as lists

### LINKED ISSUE #594: Non nullable lists do not get treated as a list
If I have the following schema:
```gql
type Book {
  name: String
  author: String
}
...
type Query {
  books(author): [Book!]!
}
```
and a resolver
```
Query: {
  books(obj, args, context, info) {
    return resolver(Book, {
      before... // some custom filtering logic
    })(obj, args.filter, context, info);
  }
}
```
Then when running the query, I get an error of `Expected Iterable, but did not find one for field Query.books`.

After debugging, the issue is coming from [this check if the type is a list](https://github.com/mickhansen/graphql-sequelize/blob/d39d0ce48ea248831480161f1aeff68f240f7b5c/src/resolver.js#L34).
as in this case, the type is `GraphQLNonNull(GraphQLList)`

M-Zuber:
Fixes #594

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -1,4 +1,4 @@
-import { GraphQLList } from 'graphql';
+import { GraphQLList, GraphQLNonNull } from 'graphql';
 import _ from 'lodash';
 import argsToFindOptions from './argsToFindOptions';
 import { isConnection, handleConnection, nodeType } from './relay';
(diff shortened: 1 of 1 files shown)


M-Zuber:
The last error _seems_ unconnected to my changes.
Is it an issue you have seen before?

PR #597, opened 2018-05-11
fix(relay.js): fix faulty cursor strategy

### LINKED ISSUE #596: sequelizeConnection will break on any string primary keys containing $
Imagine that the primary keys of a table are strings and the row at index 30 of the results has the primary key `FU$BAR`.

The cursor comes out to `base64('arrayconnection$FU$BAR$30')`.
`fromCursor` will then split at the wrong `$` and return `{id: 'FU', index: 'BAR'}` instead of `{id: 'FU$BAR', index: 30}`.
### Oops! 💩 
I don't understand why anyone would use an arbitrary (hence more error-prone) delimiting scheme when they could just `JSON.stringify([id, index])` and `JSON.parse` it back out.  JSON is hard to improve upon.

### Problem source code in `relay.js`
```
  let toCursor = function (item, index) {
    let id = item.get(item.constructor ? item.constructor.primaryKeyAttribute : item.Model.primaryKeyAttribute);
    return base64(PREFIX + id + SEPERATOR + index);
  };

  let fromCursor = function (cursor) {
    cursor = unbase64(cursor);
    cursor = cursor.substring(PREFIX.length, cursor.length);
    let [id, index] = cursor.split(SEPERATOR);
     ...
  };
```

jedwards1211:
fix #596

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -159,7 +159,7 @@ export function sequelizeConnection({
    */
   let toCursor = function (item, index) {
     let id = item.get(item.constructor ? item.constructor.primaryKeyAttribute : item.Model.primaryKeyAttribute);
-    return base64(PREFIX + id + SEPERATOR + index);
+    return base64(JSON.stringify([id, index]));
   };
 
   /**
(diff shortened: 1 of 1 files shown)


mickhansen:
Not sure why the original relay code uses a prefix: https://github.com/graphql/graphql-relay-js/blob/master/src/connection/arrayconnection.js#L149 - It should be irrelevant.

Your PR is not passing lint, travis is failing.

jedwards1211:
@mickhansen okay, I fixed the lint error

PR #600, opened 2018-05-14
[WIP] feat(relay.js): add support for bi-directional pagination

intellix:
Thought I'd push the tests that at least work to show that it works with regular relay. The rest is coming as I get more time to work on it


mickhansen:
Am i missing something or wouldn't using before with first be redundant?

mickhansen:
https://facebook.github.io/relay/graphql/connections.htm#sec-Arguments

Issue #590, opened 2018-05-06
Pagination first/last/before/after can't go both ways


intellix:
Originally I was having issues with **last**, but decided to try **first** because it's simpler to understand and found that both are broken. I haven't noticed this before now because I've just been doing infinite scrolling to append rows which means you only go one way either forwards/backwards.

Now I've got to create an actual previous/next pagination and discovered:

**Consider these 10 rows:** a, b, c, d, e, f, g, h, i, j

### ✓ first 2 (forward)
**expected:** a, b
**actual:** a, b
**expected limit:** ASC LIMIT 2
**actual limit:** ASC LIMIT 2

### ✓ first 2 after b (forward)
**expected:** c, d
**actual:** c, d
**expected limit:** ASC LIMIT 2, 2
**actual limit:** ASC LIMIT 2, 2

### ✗first 2 before c (back)
**expected:** a, b
**actual:** d, e
**expected limit:** ASC LIMIT 2
**actual limit:** ASC LIMIT 3, 2

### ✓ last 2 (forward)
**expected:** j, i
**actual:** j, i
**expected limit:** DESC LIMIT 2
**actual limit:** DESC LIMIT 2

### ✓ last 2 before i (forward)
**expected:** h, g
**actual:** h, g
**expected limit:** DESC LIMIT 2, 2
**actual limit:** DESC LIMIT 2, 2

### ✗last 2 after h (back)
**expected:** j, i
**actual:** f, e
**expected limit:** DESC LIMIT 0, 2
**actual limit:** DESC LIMIT 3, 2

Investigated somewhat and so far I found that the cursors are these: 
`base64(PREFIX + id + SEPERATOR + index)`

So in the first set, the cursors look like:
- arrayconnection$1$0
- arrayconnection$2$1
- arrayconnection$3$2
- arrayconnection$4$3

Code for the limits and offsets generated are: 
https://github.com/mickhansen/graphql-sequelize/blob/d39d0ce48ea248831480161f1aeff68f240f7b5c/src/relay.js#L217-L219
https://github.com/mickhansen/graphql-sequelize/blob/d39d0ce48ea248831480161f1aeff68f240f7b5c/src/relay.js#L265-L270

Maybe that should be (need to test):
```ts
options.offset = args.after ? startIndex + 1 : options.offset - args.limit;
```

mickhansen:
Combining first/before and last/after was probably never adequately tested as it's generally not the way relay would query it.

[...]

PR #607, opened 2018-06-05
Use windowed queries instead of OFFSET for connections when possible (and without breaking hasNextPage/hasPreviousPage)

jedwards1211:
Great success!

### Details
* Windowed queries are used except when at least one `orderBy` attribute is not a `string` (i.e. custom function or `Sequelize.literal`)
* For `first` queries, `hasNextPage` is determined by using `limit: first + 1` and seeing if the result `nodes.length > first`.  If `hasPreviousPage` is selected in the GraphQL, it will perform a second query in the opposite direction with `limit: 1` to determine the answer `hasPreviousPage`.
* For `last` queries, `hasPreviousPage` is determined by using `limit: first + 1` and seeing if the result `nodes.length > last`.  If `hasNextPage` is selected in the GraphQL, it will perform a second query in the opposite direction with `limit: 1` to determine the answer `hasNextPage`.

### Breaking Changes
* The signature of `resolveEdge` has been changed; in place of the `args` parameter it now takes an object with this shape: `{orderAttributes: Array<string>, mustUseOffset?: ?boolean}`
* The tests were using `before` to replace `order` contents with a `Sequelize.literal`.  This will not work anymore; instead users can put their custom functions or literals directly in the values for the corresponding `orderBy` enum.


jedwards1211:
@mickhansen I checked the build failures, it was more of those pesky Postgres race conditions

test/integration/relay/connection.test.js
```diff
@@ -117,26 +117,22 @@ describe('relay', function () {
...
-                  ELSE "otherDate" End ASC`);
-            } else {
-              options.order = Sequelize.literal(`
+                  ELSE "otherDate" End`
+                : `
                 CASE
                   WHEN completed = true THEN \`createdAt\`
-                  ELSE \`otherDate\` End ASC`);
-            }
+                  ELSE \`otherDate\` End`
```
janmeier: `sequelize.dialect.QueryGenerator.quoteIdentifier()`

jedwards1211: @janmeier that's good to know, thanks for the tip!

jedwards1211:
@janmeier I realized last night it would be possible to preserve the signature of `resolveEdge`.  However, rebuilding the list of `orderAttributes` from the `args` for each edge would be wasteful.  However I could use `reselect`'s `createSelector` to avoid recomputing `orderAttributes` when the `args` are the same as the previous call.  Do you think that would be worth avoiding a breaking change?

jedwards1211:
Also glad you got the Borat reference 😄

janmeier:
> Do you think that would be worth avoiding a breaking change?

The signature for resolveEdge is not formalized anywhere, so I wouldn't worry too much about it (still a major version change, but I don't think it will affect many people). I just checked through our codebase, and we only use `resolveEgde(row)` in mutations to resolve a newly added edge, so the change wouldn't affect us. I can't really think of other cases where you want to use `resolveEdge` in your own application?

jedwards1211:
@janmeier I can't think of other cases either, but I hadn't looked into how it was being used in the mutations in detail (though I did have to fix that test).  By "your codebase", do you mean in this project, or are you also using it in production projects at Snaplytics?

jedwards1211:
We might want to accomplish this using window functions for some databases: https://use-the-index-luke.com/sql/partial-results/window-functions

src/relay.js
```diff
@@ -107,6 +107,115 @@ export function nodeType(connectionType) {
...
+    });
+    $or.push(isEqualToCursor);
+  }
+
+  // include any rows that are beyond the cursor
+
+  // given ORDER BY A ASC, B DESC, C ASC, the following code would create this logic:
+  // (A > cursorValues[A]) OR
+  // (A = cursorValues[A] AND B < cursorValues[B]) OR
+  // (A = cursorValues[A] AND B = cursorValues[B] AND C > cursorValues[C])
```
jedwards1211: @janmeier I don't suppose there's a better or more idiomatic way to do this, would you know?  I haven't actually checked how well an index on (A, B, C) would support this, though I have to think it would...

jedwards1211: @janmeier I saw this syntax for MySQL at least in an article:

```sql
select
  t.id, 
  t.value,
  case (
      select count(*)
      from t t2
      where (t2.value, t2.id) <= (t.value, t.id)
    ) % 5 
    when 0 then 1 
    else 0 
  end page_boundary
from t
order by t.value, t.id
```

src/relay.js
```diff
@@ -107,6 +107,127 @@ export function nodeType(connectionType) {
...
+  postgres: true,
+  sqlite: true,
+};
+
+function tupleComparison(model, attributes, inequality, values) {
+  const {sequelize, QueryGenerator} = model;
+  const attributesStr = attributes.map(attribute => QueryGenerator.quoteIdentifier(attribute)).join(', ');
+  const escapeOptions = {context: 'SELECT'};
+  const valuesStr = attributes.map((attribute, i) =>
+    QueryGenerator.escape(values[i], model.attributes[attribute], escapeOptions)
```
jedwards1211: @janmeier am I using this right?

jedwards1211:
transient postgres failures again

janmeier:
>  By "your codebase", do you mean in this project, or are you also using it in production projects at Snaplytics?

Production at snaplytics :) (https://graphql-7.snaplytics.io/explorer)

src/relay.js
janmeier: Yep :+1:

jedwards1211:
Okay @mickhansen I made the ordering change we were discussing in #590, fixed the tests, and added a test that pages forward twice and then backward twice: https://github.com/mickhansen/graphql-sequelize/pull/607/files#diff-f784af1a4e8d657979912a0b92ef61b9R1016

So with `orderBy: [LATEST]`, which is a descending sort on task timestamp, it goes like this:

```
query: tasks(first: 3, orderBy: [LATEST])
result edges:
  T-5s
  T-10s
  T-15s
startCursor: T-5s
endCursor: T-15s

query: tasks(first: 3, after: T-15s, orderBy: [LATEST])
result edges:
  T-20s
  T-25s
  T-30s
startCursor: T-5s
endCursor: T-15s

query: tasks(first: 3, after: T-30s, orderBy: [LATEST])
result edges:
  T-35s
  T-40s
  T-45s
startCursor: T-5s
endCursor: T-15s
query: tasks(last: 3, before: T-35s, orderBy: [LATEST])
result edges:
  T-20s
  T-25s
  T-30s
query: tasks(last: 3, before: T-20s, orderBy: [LATEST])
result edges:
  T-5s
  T-10s
  T-15s
```

jedwards1211:
@mickhansen I'm still learning the vocabulary about this because I was wrong to call these "window functions".  One term I've heard for what this PR does is "keyset pagination".  Using window functions could also be helpful for connections but as they're more exotic and less consistent among SQLs I've avoided them here.

Regarding falling back to offset:
It seems like `CASE` clauses work in `ORDER BY` but not in place of an operand in an inequality
If there were a bug caused by falling back to offset here, that same bug would already occur with the current code

What do you mean by "window total query"?  I fetch in the opposite direction because, given `first: 3, after: <id: 2>`, the first SQL query is `WHERE id > 2 ORDER BY id ASC LIMIT 4` and the results look like:

```
0
1 ? is there a way to select this element in the same query?
3 <-- startCursor
4 <
6 <-- endCursor
9 <-- included in SQL result but not the GraphQL result; indicates there's another page 
10
12
```
I can't think of a way to include `WHERE id <= 2 ORDER BY id DESC LIMIT 1` (which we need to determine if the previous page exists) in the same query unless we use a `UNION ALL`.

mickhansen:
Ah yeah i didn't look at the code closely yet just your description, if you aren't actually using WINDOW ignore what i said about the window total (windows in postgres lets you get the entire row count even if you have defined offset or limit)

mickhansen:
Do you need to include the previous element? If there is a cursor there must be a previous page, atleast from the thought that you can only get cursors onces you don't the first query.

mickhansen:
@jedwards1211 I'll review this more thoroughly tomorrow. I'd still like to see the ordering change in another PR so that can be discussed on its own with all interested parties and contributors :)

jedwards1211:
> If there is a cursor there must be a previous page, atleast from the thought that you can only get cursors onces you don't the first query.

Often that will be the case, but keep in mind that all elements <= a cursor could get updated or deleted after that cursor is created.  That's why we have to make a query if we want to get the correct answer

mickhansen:
> Often that will be the case, but keep in mind that all elements <= a cursor could get updated or deleted after that cursor is created. That's why we have to make a query if we want to get the correct answer

True enough, i'd always careful introducing something that decreases performance though.

jedwards1211:
@mickhansen ah that's true, I should do it in parallel, I'll make that change when I get a chance.  I already avoid doing the second query if the relevant `hasNextPage`/`hasPreviousPage` isn't selected in the graphql query, but I should also add an option to turn off the extra query regardless (since the spec allows us to return `false` for performance reasons).

jedwards1211:
Btw it would be cool to implement pagination using window functions in databases that support it, though according to [this article](https://use-the-index-luke.com/sql/partial-results/window-functions) it doesn't perform efficiently in Postgres.  Apparently it's a valid approach with MSSQL but right now we're not testing against MSSQL so it would be quite a bit of work to implement and test that.

test/unit/relay/mutation.test.js
```diff
@@ -91,7 +91,9 @@ describe('relay', function () {
             },
             newTaskEdge: {
               type: this.viewerTaskConnection.edgeType,
-              resolve: (payload) => this.viewerTaskConnection.resolveEdge(payload.task)
+              resolve: (payload, args, context, info) => this.viewerTaskConnection.resolveEdge(
+                payload.task, -1, '', {...info, orderAttributes: [this.Task.primaryKeyAttribute]}, payload
```
mickhansen: Why is orderAttributes required here? That should be known to the connection

mickhansen: Ah nvm, it's ofcourse the requested order by from the query. Maybe it's better if it takes a value directly from the orderBy  setup?

jedwards1211: @mickhansen problem is we would have to rebuild it from `orderBy` for every single edge, that would be a lot of redundant computation

jedwards1211: I could use `createSelector` from `reselect` to make a function that returns the `orderAttributes` from `args.orderBy` in a memoized way, without unnecessary recomputation

jedwards1211:
@mickhansen 

> True enough, i'd always careful introducing something that decreases performance though.

Yeah but compared to the performance of an `OFFSET` query?

jedwards1211:
@mickhansen Okay, this now performs the secondary `hasNextPage`/`hasPreviousPage` query in parallel, and only if those fields are requested.  In addition I merged in changes from `master`.  Is there anything else you need or can we merge this?

mickhansen:
@jedwards1211 It looks solid, i need to do a full review and been kinda swamped. I'll try and set time aside to go through it with @janmeier tomorrow at our offices :)

jedwards1211:
@mickhansen that's a good point, I'll implement that.  Last night I made it handle queries like
```graphql
{
  things(first: 1000, after: "lkasdiniweraskdjf") {
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```
by using `OFFSET 999 LIMIT 2` so that it fetches the minimum information necessary to fulfill that query.  I realized people may do a query like this if their UI supports scrolling way ahead, and they want to skip over everything the user scrolled past and fetch only the rows they scrolled to.

jedwards1211:
@mickhansen actually I think we should just throw an error saying "modifying order isn't supported" if they change the order in `before` and the cursor they passed in is not in offset form.  We wouldn't be able to return a proper cursor; an attribute-based cursor might be wrong and, but we wouldn't know the starting offset to add for returning an offset-based cursor.

jedwards1211:
@mickhansen come to think of it, I'd like to add more logic to throw errors if they passed an invalid cursor.

jedwards1211:
@mickhansen I just realized there's currently a mistake: I'm passing `after` to the nodes resolver as well as calling it with the entire resolved connection.

I think I should accept a separate `afterNodes` argument.  Which makes me wonder if we should change `before` to `beforeNodes` since it's run before the nodes resolver, which may be called twice, rather than before the entire operation.

PR #618, opened 2018-06-23
fix connection pagination (without fix to return correct order for last/before)

jedwards1211:
@mickhansen  Here is my fix with the old, incorrect edges order, as you requested

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -168,6 +168,22 @@ function fromCursor(cursor) {
   return JSON.parse(unbase64(cursor));
 }
 
+const dialectsThatSupportTupleComparison = {
+  mysql: true,
+  postgres: true,
+  sqlite: true,
+};
+
+function tupleComparison(model, attributes, inequality, values) {
+  const {sequelize, QueryGenerator} = model;
+  const attributesStr = attributes.map(attribute => QueryGenerator.quoteIdentifier(attribute)).join(', ');
+  const escapeOptions = {context: 'SELECT'};
+  const valuesStr = attributes.map((attribute, i) =>
+    QueryGenerator.escape(values[i], model.attributes[attribute], escapeOptions)
+  ).join(', ');
+  return sequelize.literal(`(${attributesStr}) ${inequality} (${valuesStr})`);
+}
+
 function getWindow({model, cursor, order, inclusive}) {
   const values = fromCursor(cursor);
   order.forEach(([orderAttribute], index) => {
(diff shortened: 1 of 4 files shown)


jedwards1211:
@mickhansen @janmeier are you guys waiting for anything on this?

The test failures were as usual due to the postgres race condition.

mickhansen:
Hey @jedwards1211, sorry - have just been busy.

src/relay.js
```diff
@@ -107,6 +107,127 @@ export function nodeType(connectionType) {
...
+
+
+/**
+ * Creates a cursor given a node returned from the Database
+ * @param  {Object}   node            sequelize model instance
+ * @param  {String[]} orderAttributes  the attributes pertaining in ordering
+ * @return {String}                   The Base64 encoded cursor string
+ */
+function toCursor(node, info) {
+  return base64(JSON.stringify(info.orderAttributes.map(attr => node.get(attr))));
```
mickhansen: this makes it a migration-breaker,
any way we can make it so old cursors would still work?

src/relay.js
```diff
@@ -107,6 +107,127 @@ export function nodeType(connectionType) {
...
+  const {sequelize, QueryGenerator} = model;
+  const attributesStr = attributes.map(attribute => QueryGenerator.quoteIdentifier(attribute)).join(', ');
+  const escapeOptions = {context: 'SELECT'};
+  const valuesStr = attributes.map((attribute, i) =>
+    QueryGenerator.escape(values[i], model.attributes[attribute], escapeOptions)
+  ).join(', ');
+  return sequelize.literal(`(${attributesStr}) ${inequality} (${valuesStr})`);
+}
+
+function getWindow({model, cursor, order, inclusive}) {
```
mickhansen: can you provide an example of what SQL this would generate?

src/relay.js
jedwards1211: I'm confused what you mean, cursors are just opaque strings, and there's no way an old-style cursor would somehow pop up in the new code

src/relay.js
jedwards1211: Basically `(name, id) < ('foo', 5)` for instance

src/relay.js
jedwards1211: I mean unless users do something weird like persisting cursors until they upgrade, which we should discourage people from doing

src/relay.js
jedwards1211: Instead of `name < 'foo' OR (name = 'foo' AND id < 5)`

PR #622, opened 2018-07-17
Fix DateType -> dateType typo

RichAyotte:


RichAyotte:
@mickhansen Package is broken atm. Not sure how this didn't get caught. Maybe another test to catch this type of error would help.

mickhansen:
v9.0.1

PR #627, opened 2018-09-19
Fix issue in argsToFindOptions 

joewoodhouse:
In my database it's common for us to have a column called "order" for specifying the relative order of things.

When doing a query such as

```js
{
  objects(order: "id"){
    id,
    order
  }
}
```

when using defaultListArgs(), the resulting findOptions are currently generated as
```js
{
  where: { order: 'id' },
  order: [ [ 'id', 'ASC' ] ]
}
```

this results in the wrong query (I did not ask to filter by `order: 'id'`) and will most likely generate an SQL syntax error e.g. `invalid input syntax for integer: "id"`

The fix changes `argsToFindOptions` to handle the special cases of `offset`,`limit` and `order` first, and then assessing the targetAttributes.

### DIFF
--- a/test/unit/argsToFindOptions.test.js
+++ b/test/unit/argsToFindOptions.test.js
@@ -37,4 +37,18 @@ describe('argsToFindOptions', function () {
     expect(findOptions).to.have.ownProperty('offset');
     expect(findOptions.offset).to.be.equal(1);
   });
+
+  it('should allow filtering by "order" column when in targetAttributes', function () {
+    var findOptions = argsToFindOptions({ where: { order: 1 } });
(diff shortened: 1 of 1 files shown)


joewoodhouse:
@janmeier agreed, just pushed new tests up for that.

Any ideas what's going on with the travis-ci tests? I seem to have made them fail but I've no idea if it's due to my change. Failures like:
```
AssertionError: expected '0.7163867175254655' to equal '0.6084338268942417'
```
don't seem to relate at all to my change. I see they fail quite regularly on other branches too, are they reliable?

xsbchen:
@mickhansen can merge this pr?

PR #632, opened 2018-10-16
feat(resolver.js)Support handleConnection for root query

hank-cp:
Support connection on root Query
```graphql
type Query {
  users(
    first: Int
    after: String
    last: Int
    before: String
  ): UserConnection
}
```
or 
```js
    var userConnection = connectionDefinitions({name: 'User', nodeType: userType});

    schema = new GraphQLSchema({
      query: new GraphQLObjectType({
        name: 'RootQueryType',
        fields: {
          user: {
            type: userConnection.connectionType,
            args: connectionArgs,
            resolve: resolver(User)
          },
          node: nodeField
        }
      })
    })
```
Unit test included :)

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -76,29 +76,32 @@ function resolverFactory(targetMaybeThunk, options = {}) {
         findOptions.order = [[model.primaryKeyAttribute, 'ASC']];
       }
 
+      function checkConnection() {
+        return options.handleConnection && isConnection(info.returnType);
+      }
+
+      function handleConnectionIfNeeded(result) {
+        if (checkConnection()) {
+          return handleConnection(result, args);
+        }
+        return result;
+      }
+
       if (association) {
         if (source.get(association.as) !== undefined) {
           // The user did a manual include
           const result = source.get(association.as);
-          if (options.handleConnection && isConnection(info.returnType)) {
-            return handleConnection(result, args);
-          }
+          return handleConnectionIfNeeded(result);
 
-          return result;
         } else {
           return source[association.accessors.get](findOptions).then(function (result) {
-            if (options.handleConnection && isConnection(info.returnType)) {
-              return handleConnection(result, args);
-            }
-            return result;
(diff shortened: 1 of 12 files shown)


src/resolver.js
```diff
@@ -95,7 +95,15 @@ function resolverFactory(targetMaybeThunk, options = {}) {
...
 
-      return model[list ? 'findAll' : 'findOne'](findOptions);
+      if (options.handleConnection && isConnection(info.returnType)) {
+        return model['findAll'](findOptions).then(function (result) {
+          return handleConnection(result, args);
+        });
+
+      } else {
+        return model[list ? 'findAll' : 'findOne'](findOptions);
+      }
```
mickhansen:
```
return model[(list || options.handleConnection && isConnection(info.returnType)) ? 'findAll' : 'findOne'](findOptions).then(function (result) {
  if (options.handleConnection && isConnection(info.returnType)) return handleConnection(result, args);
  return result;
});
```

src/resolver.js
```diff
@@ -95,7 +95,12 @@ function resolverFactory(targetMaybeThunk, options = {}) {
         }
       }
 
-      return model[list ? 'findAll' : 'findOne'](findOptions);
+      // eslint-disable-next-line max-len,no-extra-parens
```
mickhansen: hmm build still failing on linting

src/resolver.js
```diff
@@ -76,26 +76,34 @@ function resolverFactory(targetMaybeThunk, options = {}) {
         findOptions.order = [[model.primaryKeyAttribute, 'ASC']];
       }
 
+      function checkConnection() {
```
mickhansen: i like this approach but defining the functions everytime seems like a waste, move them outside and have them take all args required please.

hank-cp:
Done, 1 more char triggers eslint warning...huuuu

hank-cp:
This build failed looks not caused by my changes

PR #638, opened 2018-11-06
support for Sequelize.REAL

emanuelet:


src/typeMapper.js
```diff
@@ -110,7 +113,8 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
     return returnType;
   }
 
-  if (sequelizeType instanceof JSON) {
+  if (sequelizeType instanceof JSONB ||
```
mickhansen: throw in a test for JSONB aswell

emanuelet: there is already at  https://github.com/mickhansen/graphql-sequelize/blob/f85aabe549aff265c3dd392dfbe8fe9904d9cbf3/test/unit/typeMapper.test.js#L196

mickhansen: So either the test failed before or JSONB is a subset of JSON (it probably is IIRC)

PR #642, opened 2018-11-15
Add support for passing orderBy by type name

### LINKED ISSUE #639: Relay createConnectionResolver orderBy depends on GraphQLEnumType
I'm currently in the process of trying to rewrite the JS Based schema creation methods into SDL/Resolvers.

**user.type.ts**
```ts
import gql from 'graphql-tag';

export const types = gql`
  type Query {
    user(id: ID!): User
    users(
      after: String,
      before: String,
      first: PaginationAmount,
      last: PaginationAmount,
      order: UserOrderBy
    ): UserConnection
  }

  type User {
    id: ID!
    createdAt: SequelizeDate
    updatedAt: SequelizeDate
    email: String
    name: String
  }

  type UserConnection {
    pageInfo: PageInfo!
    edges: [UserEdge]
    total: Int
  }

  type UserEdge {
    node: User
    cursor: String!
    index: Int
  }

  enum UserOrderBy {
    ID
    BALANCE
  }
`;
```

**query.resolver.ts**
```ts
export const QueryResolver = {
  Query: {
    user: (source: any, args: any, context: any, info: any) => {
      args.id = fromGlobalId(args.id).id;
      return resolver(() => User)(source, args, context, info);
    },
    users: relay.createConnectionResolver({
      target: User,
    }).resolveConnection,
  },
};
```

This is all good until I need to pass in the orderBy options (UserOrderBy specifically).
It's dependant on using JS to create that type like so:

```ts
new GraphQLEnumType({
  name: 'UserOrderBy',
  values: {
    ID: { value: ['id', 'ASC'] },
    BALANCE: { value: ['balance', 'ASC'] },
  },
});
```

https://github.com/mickhansen/graphql-sequelize/blob/b036ef39b4831f682e62edd6654fcf7b4bea92ab/src/relay.js#L207-L209

intellix:
Partially resolves #639 

I thought we could resolve it by just allowing either String or EnumType but let me know if you'd prefer a separate parameter. I could document this Behaviour in relay.md

### DIFF
--- a/docs/relay.md
+++ b/docs/relay.md
@@ -177,7 +177,7 @@ fragment getCreated on userTaskEdge {
 ```
 
 You can pass custom args in your connection definition and they will
-automaticly be turned into where arguments. These can be further modified
+automatically be turned into where arguments. These can be further modified
 using the `where` option in `createConnection`.
 
 ```js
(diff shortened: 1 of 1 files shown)


intellix:
Strange that it passed locally and can see 10 passed but the other errors for Node 6/8 don't seem to make sense to me. Any idea?

PR #643, opened 2018-11-15
Add support for string based type names in nodeTypeMapper

intellix:
The rest of #639


mickhansen:
`v9.2.0` - thanks @intellix

Issue #611, opened 2018-06-08
MSMQL Support?


jon64digital:
I've been looking for a way to use MS SQl Serverwith Node/GraphQL for a while avoiding the N+1 query problem and thought I'd hit the jackpot here. I see that it's not currently supported though. Is this because it's not possible to support it or is there something planned?

mickhansen:
It's mostly because none of the people involved work with MSSQL

jon64digital:
Ah, OK, so it is possible then. Would be great to add it as Sql Server is one of the most used databases in the world, especially in enterprise.

UmerIftikhar:
Is MS SQL support still to be added?

mickhansen:
@UmerIftikhar Not by me.

jon64digital:
I'm still hoping that somebody watching this does have the knowledge to implement this. MSSQL is only behind Oracle and MySql in terms of being the most popular database in the world looking at recent stats so I'm sure there must be way more people out there that need this compared to something like Postgresql support.

mickhansen:
@jon64digital it might be very popular globally but it doesn't seem to have gained much traction in Nodeland.

gvbkr:
@mickhansen  Quick question - what does it take to add support for MS SQL Server? Since we are using sequelize shouldn't the actual database be abstracted? I can help make the changes if you can help give us an overview of what needs to be done?

@jon64digital  @UmerIftikhar  may be we can work together to add this support once we have clarity on what needs to be done.

mickhansen:
@gvbkr Sequelize doesn't always have the exact same support for all dialects. But this "fix" sort of relies on a bug. Sequelize needs full cross dialect support for `offset` in `groupedLimit`, then we can rely on that here.

mickhansen:
@gvbkr I don't know, for the most part it should just work but there can be subtle differences.
If a few people are willing to help maintain mssql support i'm open to that.

mubaidr:
Looks like this does work fine with MSSQL except that it does not seems to support some data types. It emits error when encounter `NCHAR` or `SMALLDATETIME` data type. I did modify the code to treat these as string and it started working. 

I suggest all non-existence dataTypes should be treated as string by default.

mickhansen:
> I suggest all non-existence dataTypes should be treated as string by default.

I'd rather throw, defaulting to string means breaking that when we fix it (if string is incorrect). Users can ignore certain fields and provide their own types or submit a PR :)

[...]

PR #644, opened 2018-11-16
Fix for missing Data Types for MSSQL

mubaidr:
'NCHAR', 'SMALLDATETIME' Data Types should be treated as Strings

### DIFF
--- a/src/typeMapper.js
+++ b/src/typeMapper.js
@@ -83,7 +83,8 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
       sequelizeType instanceof DATEONLY ||
       sequelizeType instanceof TIME ||
       sequelizeType instanceof BIGINT ||
-      sequelizeType instanceof DECIMAL) {
+      sequelizeType instanceof DECIMAL ||
+      ['NCHAR', 'SMALLDATETIME'].indexOf(sequelizeType) > -1) {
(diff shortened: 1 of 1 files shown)


mubaidr:
This needs to be handled in Sequelize implementation only then we rewrite this pull request.

src/typeMapper.js
```diff
@@ -83,7 +83,8 @@ export function toGraphQL(sequelizeType, sequelizeTypes) {
       sequelizeType instanceof DATEONLY ||
       sequelizeType instanceof TIME ||
       sequelizeType instanceof BIGINT ||
-      sequelizeType instanceof DECIMAL) {
+      sequelizeType instanceof DECIMAL ||
+      ['NCHAR', 'SMALLDATETIME'].indexOf(sequelizeType) > -1) {
```
mickhansen: defined as strings in your sequelize model?

mubaidr: Sequelize does not handle these types and therefore `sequelizeType` is just a string in this case. This thing needs to be handled in Sequelize. Closing this PR.

mubaidr: But do need to fix this issue in mentioned in comment: https://github.com/mickhansen/graphql-sequelize/issues/611#issuecomment-439342438

Issue #646, opened 2018-12-11
Cache enum types in attributeFields doesn't work as expected


alirizwan:
When calling attribute fields multiple times on same model, duplicate enumtypes gets created with an error:
`Error: Schema must contain unique named types but contains multiple types named "UserroleEnumType".`

mcouthon:
Why was this closed? This still seems like an existing bug.

mcouthon:
Actually, I solved it by doing this:
```javascript
const options = {
    cache: {},
    commentToDescription: true
  };
const fullModel = attributeFields(model, options);
options.allowNull = true;
const looseModel = attributeFields(model, options);
}
```

Seems to work well

PR #647, opened 2018-12-11
cache always initializes to {} and it doesn't cache enum types when a…

alirizwan:
…ttributeFields called multiple times on same model.


alirizwan:
related to issue https://github.com/mickhansen/graphql-sequelize/issues/646

PR #650, opened 2018-12-23
implemented checkDefaults option

Yehonal:
it allows attributeFields to skip  GraphQLNonNull wrapping when default value exists on sequelize model
it can be used to generate graphql input types


Schlesiger:
+1 This would be quite handy

PR #657, opened 2019-02-21
adding sql lite for the counts

corbanb:
allow counts for sqlite queries. 

this will help with test runners using sqlite

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -247,7 +247,7 @@ export function createConnectionResolver({
             model.sequelize.literal('COUNT(*) OVER()'),
             'full_count'
           ]);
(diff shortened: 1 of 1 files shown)


corbanb:
anyone know what I need to do for the tests to use sqlite? also not sure why this would even fail if its not running sqlite in the test suite 👎

mickhansen:
Failure is unrelated, there's a wonky non-deterministic test that fails here and there.

mickhansen:
Ill merge this but just keep in mind this library does not support sqlite specifically.

mickhansen:
v9.3.4

PR #661, opened 2019-03-21
Ignore docker configs and git pre-commit hook

jonmagic:
While debugging package installation issues I noticed that there were some things not in `.npmignore` that should probably be added:

```
Dockerfile
docker-compose.yml
pre-commit
```


mickhansen:
Thanks @jonmagic

PR #662, opened 2019-03-21
Use Buffer.from

jonmagic:
I was receiving deprecation notices when using with Node 10: https://nodesource.com/blog/understanding-the-buffer-deprecation-in-node-js-10/

### DIFF
--- a/src/base64.js
+++ b/src/base64.js
@@ -1,9 +1,9 @@
 'use strict';
 
 export function base64(i) {
-  return (new Buffer(i, 'ascii')).toString('base64');
+  return (new Buffer.from(i, 'ascii')).toString('base64');
 }
 
 export function unbase64(i) {
-  return (new Buffer(i, 'base64')).toString('ascii');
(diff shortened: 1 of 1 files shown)


src/base64.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
 export function base64(i) {
-  return (new Buffer(i, 'ascii')).toString('base64');
+  return (new Buffer.from(i, 'ascii')).toString('base64');
 }
 
 export function unbase64(i) {
-  return (new Buffer(i, 'base64')).toString('ascii');
+  return (new Buffer.from(i, 'base64')).toString('ascii');
```
mickhansen: As @intellix mentions `new` can be removed

jonmagic: Yeah, that was a whoops on my part.

jonmagic:
Is this a flaky test or did I actually break something with this change? Not sure how to re-run CI to find out.

PR #664, opened 2019-05-08
Replace `_.assign` on `Object.assign`

shvetsovdm:
This change is aimed to remove ambiguity.

Use `Object.assign` in Readme in examples because not everyone knows about lodash package, someone knows that `_` is underscore package, `_` is not required/imported explicitly.


mickhansen:
Thanks @shvetsovdm

Issue #665, opened 2019-05-08
Where operators not being replaced when using relay resolver


Azerothian:
with fix #624  we stopped mutating the args.where object, 

this in turn has broken the argsToWhere function in the before event on the relay resolver. 

As it processes the original args.where object and wipes out the previously converted options.where. leaving the operators unconverted and causes an error for any complex query. 

e.g. 
`{
  "name": {
    "iLike": "someone"
  }
}`

https://github.com/mickhansen/graphql-sequelize/blob/master/src/relay.js#L258

Azerothian:
thank you

PR #666, opened 2019-05-08
#665 - fix for relay where operators

Azerothian:
#665 - fix for relay where operators


mickhansen:
`v9.3.6`

Thanks @Azerothian

Issue #670, opened 2019-06-28
Problem setting dataloader contextToOptions with dataloader-sequelize 2.0.0


caioflores:
Following this discussion: https://github.com/mickhansen/graphql-sequelize/issues/656

The example server of dataloader with graph-yoga is not working anymore with dataloader-sequelize 2.0.0: https://github.com/mickhansen/graphql-sequelize/blob/master/examples/graphql-yoga/src/server.js

The new way to make it work as @intellix solved is:

```
schema.context = req => {
  const dataloaderContext = createContext(sequelize);

  resolver.contextToOptions = {
    dataloaderContext: [EXPECTED_OPTIONS_KEY]
  };
  return {
    ...req,
    dataloaderContext
  };
};
```

I'll open a PR to update README and example.

caioflores:
Solved the problem here #671

mickhansen:
Will discuss in PR.

PR #671, opened 2019-06-28
update to dataloader-sequelize 2.0.0

caioflores:
- Update dataloader-sequelize to 2.0.0 version.
- Update server to handle contextToOptions in the correct way.
- Update resolvers to pass context do resolver.

Solves #670


examples/graphql-yoga/src/server.js
```diff
@@ -26,16 +26,16 @@ const typeDefs = `
...
 const resolvers = {
   Query: {
-    pet: resolver(models.Pet),
-    pets: resolver(models.Pet),
-    user: resolver(models.User),
-    users: resolver(models.User),
+    pet: (...args) => resolver(models.Pet)(...args),
+    pets: (...args) => resolver(models.Pet)(...args),
+    user: (...args) => resolver(models.User)(...args),
+    users: (...args) => resolver(models.User)(...args),
```
mickhansen: should not be required, unless `graphql-yoga` changed something i guess.

examples/graphql-yoga/src/server.js
```diff
@@ -52,8 +52,12 @@ const server = new GraphQLServer({
 
     // Using the same EXPECTED_OPTIONS_KEY, store the DataLoader context
     // in the global request context
+    resolver.contextToOptions = {
+      dataloaderContext: [EXPECTED_OPTIONS_KEY]
+    };
+
     return {
-      [EXPECTED_OPTIONS_KEY]: dataloaderContext,
+      dataloaderContext,
```
mickhansen: old way would have worked just fine as long as context maps `EXPECTED_OPTIONS_KEY` to `EXPECTED_OPTIONS_KEY`

examples/graphql-yoga/src/server.js
caioflores: I tested here, it's not working without it. I lost so much time trying to debug it rsrs.

examples/graphql-yoga/src/server.js
caioflores: Yeap, you're right, just updating the resolvers I got the correct result.

examples/graphql-yoga/src/server.js
caioflores: Wait, you are saying that it's not necessary in Query context, but it's necessary in: 

```
  User: {
    pets: (...args) => resolver(models.User.Pets)(...args),
  },
  Pet: {
    owner: (...args) => resolver(models.Pet.Owner)(...args),
  },
```
?

mickhansen: I never personally used `graphql-yoga` but assuming you use the same version as this example they must have made a breaking change.

However, in any case, `users: (...args) => resolver(models.User)(...args)` should be functionally equivelant to `users: resolver(models.User)`. Unless perhaps graphql-yoga checks the function length of the resolver and does something weird.

examples/graphql-yoga/src/server.js
caioflores: I removed this change, since it was not necessary.

examples/graphql-yoga/src/server.js
caioflores: My tests here, you can see that the batch just work when I use (...args)

teste1
teste2

intellix: Isn't this required because the resolver context is now request based and not server based? Had to do this because you need to set it on each request

caioflores: @intellix maybe, I'll search about it.

mickhansen: @intellix No the resolver factory should still look at the context for the request, the resolver factory returns it's own method so with this code pattern you just keep recreating the function

caioflores: @mickhansen I found something here. First, I was messing up because I was using yarn, so I fixed it and made some more tests. 

I was testing here how to reproduce the problem, I discovered that the problem is related to dataloader-sequelize 2.0.0. When I update only the graphql-yoga version to 1.8.0 it works without this (...args), but when I update only dataloader to 2.0.0 it doesn't work and it consoles me this warning (no original function findByPk to wrap).

So maybe this is related to this dataloader-sequelize update to version 2.0.0.

hamzakubba: FWIW, I'm not using `graphql-yoga` and I did need:
```
resolve: (...args) => resolver(relationshipReference)(...args)
```
rather than
```
resolve: resolver(relationshipReference)
```
which did not work.

Relevant package versions:
```
    "dataloader": "^2.0.0",
    "dataloader-sequelize": "^2.3.1",
    "express-graphql": "^0.12.0",
    "graphql": "^15.5.0",
    "graphql-relay": "^0.6.0", // <--- not sure why this was required, but it was once I added dataloader-sequelize
    "graphql-sequelize": "^9.4.3",
```

PR #672, opened 2019-06-28
Add title linking to graphql-yoga example

caioflores:
It took me some time to find the graphql-yoga example, I added it to README to help others.


caioflores:
I only updated the README, why this broke a test?

PR #674, opened 2019-07-02
fix: add default scope order

### LINKED ISSUE #488: Resolver overwrites defaultScope order
Seeing an issue with the resolver overwriting a default order set in the model using `defaultScope`.  Here is an example:

```
'use strict'
module.exports = function (sequelize, DataTypes) {
  var Room = sequelize.define('room', {
    rank: {
      type: DataTypes.INTEGER,
      defaultValue: 0,
      allowNull: false
    }
  }, {
    defaultScope: {
      order: [['rank', 'ASC']]
    }
  })
  return Room
}
```
And the query:
```
rooms: {
  type: new GraphQLList(new GraphQLNonNull(roomType)),
  args: defaultListArgs(Room),
  resolve: resolver(Room)
}
```

The SQL output is `SELECT "id", "name", "description", "sensorId", "rank", "createdAt", "updatedAt", "storeId" FROM "rooms" AS "room" ORDER BY "room"."id" ASC;`

If I run a `findAll` query manually, I get it ordered by rank as expected:
`Executing (default): SELECT "id", "name", "description", "sensorId", "rank", "createdAt", "updatedAt", "storeId" FROM "rooms" AS "room" ORDER BY "room"."rank" ASC;`

I believe this is due to the following lines in the resolver:
https://github.com/mickhansen/graphql-sequelize/blob/master/src/resolver.js#L50-L52

Jekiwijaya:
Current behavior override the defaultScope order, this commit fix by taking the defaultScope order first then add the default order by id.

Closed #488

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -86,7 +86,7 @@ function resolverFactory(targetMaybeThunk, options = {}) {
       }
 
       if (list && !findOptions.order) {
-        findOptions.order = [[model.primaryKeyAttribute, 'ASC']];
+        findOptions.order = model._scope.order || [[model.primaryKeyAttribute, 'ASC']];
       }
 
       if (association) {


mickhansen:
PR still looks good, just needs a test

PR #682, opened 2019-09-15
Support for latest sequelize version

nikhildaga:
node query with latest sequelize version results in `sequelize.models[model].findById is not a function` error.

This pr fixes it.


mickhansen:
That fix will break other versions, please provide a fix that will support multiple versions of sequelize.

PR #685, opened 2019-11-01
Update README.md

tonmanayo:
As of later versions of sequilize you have to use sequilize.op 
https://sequelize.org/master/manual/querying.html


README.md
```diff
@@ -96,6 +96,9 @@ about those is available in their [resolver docs](http://graphql.org/learn/execu
 
 ```js
 import {resolver} from 'graphql-sequelize';
+const Sequelize = require('sequelize');
```
mickhansen: es6 import like the rest of the readme, might aswell

Issue #356, opened 2016-10-26
Order by multiple columns


manhhailua:
I have just realized that the `order` arguments which added by `defaultListArgs` can not be used for multiple columns (or I have not known how to do).

Expected syntax of variables:

```
{
  "order": [
    "createdAt",
    "reverse:updatedAt"
  ]
}
```

or as Sequelize default:

```
{
  "order": [
    ["createdAt", "ASC"],
    ["updatedAt", "DESC"]
  ]
}
```

mickhansen:
Instead of relying on defaultListArgs i'd suggest you instead use an ENUM for defining order by.

tobyhede:
@manhhailua I have a patch that does this if you need it https://github.com/mickhansen/graphql-sequelize/pull/438/

iDVB:
+1 Is there a merged solution for this yet?

cbravo:
My simple solution was to leverage the before option in the resolver as shown in the resolver tests (https://github.com/mickhansen/graphql-sequelize/blob/master/test/integration/resolver.test.js):

`import myModel from '../models/myModel';
 
 const listThings = {
   type: new ListType(thingsType),
   args: Object.assign(defaultArgs(myModel), defaultListArgs(), {
     orderBy: {
       type: new ListType(new ListType(StringType)),
     },
   }),
   description: 'Queries the things',
   resolve: resolver(myModel, {
     before: (opts, args) => {
       const options = Object.assign({ order: [] }, opts);
       if (args.orderBy) {
         options.order = options.order.concat(args.orderBy);
       }
       return options;
     },
   }),
 };
 
 export default listThings;`


and your query should look something like:

`{
  listThings(orderBy:[["field1", "DESC"], ["field2", "DESC"]]) {
    field1,
    field2,
    field3,
    field4
  }
}`

manhhailua:
@cbravo : this can be a solution for now but I keep my opinion about improving current "order".
@mickhansen : there are current interests in this issue so I will keep myself on this. I hope another PR will be better my last one.

pbrink231:
This would be great to improve the default order to include this.

I have a lot of tables and this would make working with tables server side easy.  I saw the reverse addition you made but really having an easy way to put asc or desc in the order field would be a much better way.  This also solves ordering by multiple columns in the way you think it should happen.

I think to make a non breaking change is to add orderBy as a default along with order.  You can either send in just the field name with order or you can do a more indepth ordering with orderBy.

[...]

PR #693, opened 2020-01-14
Implemented support for multiple fields on orderBy

Yehonal:
It will fix: https://github.com/mickhansen/graphql-sequelize/issues/356

### DIFF
--- a/src/argsToFindOptions.js
+++ b/src/argsToFindOptions.js
@@ -11,11 +11,15 @@ export default function argsToFindOptions(args, targetAttributes) {
         } else if (key === 'offset') {
           result.offset = parseInt(args[key], 10);
         } else if (key === 'order') {
-          if (args[key].indexOf('reverse:') === 0) {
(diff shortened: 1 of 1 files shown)


mickhansen:
As i commented in the original thread i would suggest people rely on using something more graphql idiomatic (enums) than sending in arbitrary strings.

Yehonal:
> As i commented in the original thread i would suggest people rely on using something more graphql idiomatic (enums) than sending in arbitrary strings.

This imply to change the type of order argument, is it right?

This will break compatibility with old code, instead such change is legacy and just allows you to do the following:

"reverse:field1,reverse:field2,field3" etc...

Keeping the current syntax. 
Then in ver 9.5 or 10 you can change this behaviour.

What do you think?

Yehonal:
PS travis error are not related to this change, I guess

```
 AssertionError: expected '0.18102374714205038' to equal '0.9222949639571385'
      + expected - actual
      -0.18102374714205038
      +0.9222949639571385
```

PR #699, opened 2021-02-12
Add CITEXT to typeMapper

AllAwesome497:
Add `CITEXT` (case insensitive string) since it isn't present here.


mickhansen:
v9.4.3

PR #711, opened 2022-07-01
fix order

alichherawalla:
Hi, 

Thank you for this amazing library. We use it extensively and has helped solve a lot of our problems and increase velocity. 

However we identified that order doesn't work when using sequelize instead of dataloader-sequelize. Adding in this change fixes it, else it always defaults to ["id", "asc"] for the order.

### DIFF
--- a/src/relay.js
+++ b/src/relay.js
@@ -211,36 +211,39 @@ export function createConnectionResolver({
         options.limit = parseInt(args.first || args.last, 10);
       }
 
-      // Grab enum type by name if it's a string
-      orderByEnum = typeof orderByEnum === 'string' ? info.schema.getType(orderByEnum) : orderByEnum;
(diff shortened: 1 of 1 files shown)


mickhansen:
How does your code look that you are passing `options.order`? To my knowledge `createConnectionResolver` / `createConnection` do not take these, are you modifying in a hook?

alichherawalla:
Hi, apologies for the delay in response. 

I get order since I use `defaultListArgs`

alichherawalla:
Hi @mickhansen, 

Possible to please get a review?

mickhansen:
HI @alichherawalla, sorry about the delay, could you add a test to guard against regression?

alichherawalla:
Hi, 

Just saw this. Yes will add it in @mickhansen

